PR #5, opened 2025-08-01
Build for amd64 and arm64

andersonshatch:
Would be nice to have an arm image published for Apple Silicon usage.
I'm not sure if this will work as is with the build then push approach used -- presumably because of signing things -- but hard to test.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -45,6 +45,9 @@ jobs:
         
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
         
       - name: Extract metadata for Docker
         id: meta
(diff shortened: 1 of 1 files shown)


jason-at-contrast:
When you do a build for the container, you can do this: 
`docker build --platform linux/arm64/v8 -t contrast/mcp-contrast .`
Which removes the warning that you see in docker and also builds it for the latest silicon chip "v8"

andersonshatch:
Yes, we should do that in the pipeline as folks will pull the public image to avoid building. Hence this PR.

andersonshatch:
Who can merge this? The checks failing are surely unrelated, transient, or need fixing elsewhere.

ChrisEdwards:
@andersonshatch Because this PR is coming from your fork of this repo, I can't rerun the build. I think that is why those checks are failing. I'm going to see of the other devs think its safe to merge otherwise.

JacobMagesHaskinsContrast (approved):

dougj-contrast:
> @andersonshatch Because this PR is coming from your fork of this repo, I can't rerun the build. I think that is why those checks are failing. I'm going to see of the other devs think its safe to merge otherwise.

This is fine to merge. The check failures are failing due to missing secrets in the fork repo (I believe). Once merged they will run on the default branch and we can validate they pass.

ChrisEdwards:
Merging since everyone approves.

PR #8, opened 2025-08-08
Hostname protocol fix

JoeBeeContrast:
This fix allows the user to specify the TS server as either 
example.contrastecurity.com or https://example.contrastsecurity.com
If the protocol is set in the config, that is used. If not it defaults to what is in the application.properties. And if not there for some reason defaults to https.


jason-at-contrast (commented):
This looks solid.  I'll build and test it now.

jason-at-contrast (approved):
Tested and works, no issues.

PR #11, opened 2025-08-19
Fix JSON parsing error for missingRequiredFields

ChrisEdwards:
Changes the type of missingRequiredFields from List<String> to List<ValidationErrorFields> to match the actual JSON structure returned by the TeamServer API.

Fixes error: Expected a string but was BEGIN_OBJECT at line 1219 column 34 path $.applications[19].missingRequiredFields[0]

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,83 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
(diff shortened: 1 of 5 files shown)


jason-at-contrast (approved):
So, looking back on what I originally called that object, "validationErrorFields" and we probably should refactor it and call it "Field" and then the list of fields makes more sense... other than that, the code looks solid with the reuse...

PR #13, opened 2025-08-25
Feature/risk tolerance risk level

jason-at-contrast:
building in some safeguards so if/when a customer decides to hook this up to their contrast data feed, they are aware of some risk that come with it.  The formal acceptance risk tolerance will need to be set in the configuration along with the contrast configurations.  The documentation is also updated to reflect these changes.  This is a phase 1 approach to get something in the works for protection right now.  big advancements around the configuration of these functions and what you may want to set the risk settings to be.  For now, feel free to clone/change those risk settings, rebuild/deploy if needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,7 @@ build/
 
 ### VS Code ###
 .vscode/
+
+### contrast specific items ###
+config.json
+*.ipynb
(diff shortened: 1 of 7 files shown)


jason-at-contrast:
> No issues from a secure coding perspective. You might want to run this through a quality checker and linter as well. Did you want to write some unit tests for this as well? Test coverage is pretty low at the moment.

We should setup a quality checker and linter in the pipeline.  I'll look at getting those added, probably not part of this PR though.  and Yes i agree, the whole project is low on test coverage.  I'll add that as a thing to do for the whole project.

PR #85, opened 2026-03-17
AIML-549: Refactor: Replace mutable warnings List with WarningCollector

ChrisEdwards:
## Why This Change Exists

The base classes `PaginatedTool` and `SingleTool` threaded a raw `ArrayList<String> warnings` as a mutable parameter through the entire call stack. Any method at any depth could call `.clear()`, `.remove()`, or `.set()` — nothing in the type enforced the intended contract of append-only accumulation. The contract was invisible and unenforced.

`GetVulnerabilityTool` exposed the most visible pain point: the same 8-line try/catch/warn pattern repeated three times for optional data fetches:

```java
try {
    var recommendationResponse = sdk.getRecommendation(orgId, vulnId);
    if (recommendationResponse != null && recommendationResponse.getRecommendation() != null) {
        recommendationText = recommendationResponse.getRecommendation().getText();
    }
} catch (Exception e) {
    log.debug(\"Could not fetch recommendation for {}: {}\", vulnId, e.getMessage());
    warnings.add(\"Recommendation data not available: \" + e.getMessage());
}
```

Beyond the boilerplate, several correctness problems existed:

- **Exception messages leaked to agents** — `ListApplicationsByCveTool` included `e.getMessage()` directly in agent-visible warnings, potentially exposing internal API paths, hostnames, or authentication hints
- **HttpResponseException in base classes lost accumulated warnings** — `handleHttpResponseException` called `SingleToolResponse.error(msg)` which discards any warnings collected before the exception
- **Missing diagnostic context for agents** — `GetRouteCoverageTool` used `log.warn(...)` for null-check branches but never called `collector.warn(...)`, so agents saw a generic \"Resource not found\" with no detail about which resource
- **Non-deterministic warning format** — some sites used `": " + e.getMessage()`, others used `"not available"`, with no `(retrieval error)` signal for agents to distinguish transient failures from legitimate absence

---

## Approach We Chose

Introduce `WarningCollector` — a plain `final` class (not a Spring bean) that encapsulates the try/catch/warn pattern as a first-class method.

The base classes construct one instance per request and pass it into `doExecute()` in place of `List<String> warnings`. Tool subclasses call `collector.tryFetch(description, supplier)` and `collector.tryRun(description, runnable)` for optional data fetches; failures are handled consistently with no boilerplate. The `snapshot()` method is package-private, so only the base classes can read warnings when building the response.

**Key design decisions:**
- `tryFetch` returns `Optional<T>` — null from the supplier is treated as absent with no warning, exception produces a warning
- `tryRun` returns `boolean` — for void operations like building stack trace data
- The `(retrieval error)` suffix in warnings is standardized and signals transient fetch failure to agents
- `HttpResponseException` includes the HTTP status code in the suffix: `"... not available (retrieval error, HTTP 503)"`
- Exception details are logged server-side but never flow into agent-visible warnings — information security boundary
- `WarningCollector` carries its own `@Slf4j` logger, removing the `Logger` constructor parameter from `forContext()`

---

## Outcome

The core refactor eliminates ~50 lines of repeated try/catch/warn boilerplate across `GetVulnerabilityTool` and `ListApplicationsByCveTool`. Every optional data fetch now uses a single, consistent pattern. Exception messages can no longer accidentally reach agent-visible output. Accumulated warnings are preserved through `HttpResponseException` paths.

All 652 unit tests pass. 19 commits from the working session landed cleanly.

---

### Step 1 — Core refactor: introduce WarningCollector

The first change introduced `WarningCollector` as a new class in `tool/base/` and updated both base classes to use it. `PaginatedTool.executePipeline` and `SingleTool.executePipeline` now construct a `WarningCollector.forContext(Map.of("requestId", requestId))` instead of `new ArrayList<>()`. The `doExecute` abstract method signatures changed from `List<String> warnings` to `WarningCollector collector`.

All 12 tool subclasses with simple `doExecute` implementations were mechanically updated — they had been calling `warnings.add(...)` which became `collector.warn(...)`. `GetVulnerabilityTool` got the most meaningful refactor: its three repeated try/catch blocks collapsed to one `tryFetch` chain each plus a single `tryRun` for stack trace assembly.

### Step 2 — Fix: exception messages leaked to agents

`ListApplicationsByCveTool.enrichAppsWithClassUsage` had a manual catch block that embedded `e.getMessage()` directly into the warning string. This was the only site in the codebase bypassing the security boundary that `WarningCollector` enforces automatically. Converting `enrichAppsWithClassUsage` to use `collector.tryRun(...)` fixed the leak and aligned the warning format with every other enrichment path.

### Step 3 — Fix: warnings lost on HttpResponseException

`SingleTool.handleHttpResponseException` called `SingleToolResponse.error(errorMessage)`, which creates a response with an empty warnings list — discarding any warnings the tool accumulated before the exception. The fix passes `collector` into the handler and calls `collector.snapshot()` when constructing the response, so previously collected warnings are preserved alongside the error message. The same fix was applied to `PaginatedTool`.

### Step 4 — Fix: WarningCollector WARN logs attributed to base class, not tool subclass

`WarningCollector.forContext(log, context)` originally accepted a `Logger` parameter from the base class, meaning all WARN entries showed `PaginatedTool` or `SingleTool` as the source in logs — not the actual tool subclass. One approach (adding `getLogger()` to `BaseTool`) was implemented and then reversed in favor of a simpler solution: give `WarningCollector` its own `@Slf4j` logger and remove the `Logger` parameter entirely. Trade-off: WARN entries now show `WarningCollector` as the source class rather than the tool subclass; the `requestId` context key still allows per-request log correlation.

### Step 5 — Fix: missing agent-visible warnings in GetRouteCoverageTool

`GetRouteCoverageTool.doExecute` had two early-return null checks that called `log.warn(...)` but not `collector.warn(...)`. When `doExecute` returned null, `SingleTool` converted it to a generic `"Resource not found"` response. Agents had no signal about which resource was missing. Adding `collector.warn(\"Session metadata not available for this application\")` and `collector.warn(\"Agent session not available for this application\")` before the null returns gives agents actionable diagnostic context.

### Step 6 — Fix: internal scan UUID in agent-visible warning

`GetSastResultsTool` included the raw scan UUID in a warning message when a scan was not found. UUIDs are internal implementation details with no agent utility. The warning was simplified to use only the project name, matching the format of the adjacent \"no completed scans\" warning.

### Step 7 — Fix: @Tool description accuracy

Two `@Tool` descriptions had stale text following the refactor:
- `get_vulnerability` said \"Warnings will indicate which optional data couldn't be retrieved\" — updated to clarify that warnings only fire on retrieval errors (transient failures), not on legitimately absent data
- `search_vulnerabilities` referenced a non-existent `message` field in the description of the response structure — removed

### Step 8 — Build toolchain fix

Running `make check-test` after Spotless touched file timestamps caused `@Slf4j`-generated `log` fields to go missing during incremental compilation. The fix replaces `useIncrementalCompilation=false` (Spotless plugin config that doesn't exist) with `annotationProcessorPaths` in the maven-compiler-plugin configuration, ensuring Lombok's annotation processor runs correctly on every incremental recompile.

---

## What Changed

| File | Change |
|------|--------|
| `tool/base/WarningCollector.java` | **New** — encapsulates try/catch/warn pattern, `tryFetch`, `tryRun`, `warn`, `snapshot()` |
| `tool/base/PaginatedTool.java` | Replace `ArrayList<String> warnings` with `WarningCollector collector`; preserve warnings on HttpResponseException |
| `tool/base/SingleTool.java` | Same as PaginatedTool; preserve warnings on HttpResponseException and UnauthorizedException |
| `tool/vulnerability/GetVulnerabilityTool.java` | Collapse 3 try/catch blocks to `tryFetch`/`tryRun` calls; remove `warnings` param from helpers |
| `tool/library/ListApplicationsByCveTool.java` | Convert `enrichAppsWithClassUsage` to `collector.tryRun`; fix exception message leak |
| `tool/coverage/GetRouteCoverageTool.java` | Add `collector.warn(...)` before null-return paths |
| `tool/sast/GetSastResultsTool.java` | Remove internal scan UUID from agent-visible warning |
| All other tool subclasses | Mechanical update: `List<String> warnings` → `WarningCollector collector`, `warnings.add` → `collector.warn` |
| `pom.xml` | Fix Lombok annotation processor for incremental builds via `annotationProcessorPaths` |
| `CLAUDE.md` | Document correct `br comments add` syntax; require `make verify` before creating PR |

---

## Test Coverage

**New test class:** `WarningCollectorTest.java` — 14 tests covering:
- `tryFetch` returns value on success, empty on null, empty + warning on exception
- `tryRun` returns true on success, false + warning on exception
- Warning format: standard `(retrieval error)` suffix
- HTTP status code in suffix when `HttpResponseException` thrown: `(retrieval error, HTTP 503)`
- Multiple failed fetches accumulate independently (no accidental clear)
- `warn()` appends unconditionally, throws on null, silently skips blank
- `snapshot()` returns immutable copy — mutation of returned list doesn't affect collector

**Updated tests:** `PaginatedToolTest`, `SingleToolTest`, `GetRouteCoverageToolTest`, `ListApplicationsByCveToolTest`, `GetSastResultsToolTest`

**Results:** 652 unit tests, 0 failures

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -318,6 +318,14 @@ Never use `$()` substitution to pass multi-line comment text — use `-f` instea
 - `in_progress` → Actively working on task (SET THIS WHEN YOU START!)
 - `closed` → Task complete, tested, and merged
 
+### Adding Comments to Beads
+
+```bash
+br comments add <bead-id> "comment text"
+```
+
+Note: `br comment` (no s) does NOT work — the correct subcommand is `br comments add`.
+
 ### Human Review Required Label
 
 **IMPORTANT: Beads labeled `needs-human-review` require human approval before AI work begins.**
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -32,6 +32,8 @@ make test VERBOSE=1
 make check VERBOSE=1
 ```
 
+**After a compilation failure**, stale `.class` files remain and cause "Unresolved compilation problems" on the next test run. Always run `mvn clean test-compile` (or `make clean && make test`) to recover before continuing.
+
 **Direct Maven commands** (verbose output, use make targets above for quiet output):
 - **Build**: `mvn clean install` or `./mvnw clean install`
 - **Test (unit)**: `mvn test`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -482,7 +482,7 @@ NOTE: This is not for parent-child dependencies, these are blocks dependencies.
    - If credentials unavailable, verify integration tests pass in CI/CD
 4. **Verify new tests are included** - Ensure your tests ran and passed
 
-All code changes require corresponding test coverage. Do not move to review without tests.
+All code changes require corresponding test coverage. Do not create a PR until `make verify` passes. Do not move to review without tests.
 
 See INTEGRATION_TESTS.md for integration test setup and credentials.
 
--- a/pom.xml
+++ b/pom.xml
@@ -109,6 +109,16 @@
 		</testResources>
 		<plugins>
 			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<!-- Spotless touches file timestamps during apply, which causes Maven's
+					     incremental compiler to recompile files without re-running Lombok's
+					     annotation processor, breaking @Slf4j log field generation. -->
+					<useIncrementalCompilation>false</useIncrementalCompilation>
+				</configuration>
+			</plugin>
+			<plugin>
 				<groupId>org.springframework.boot</groupId>
 				<artifactId>spring-boot-maven-plugin</artifactId>
 				<executions>
--- a/pom.xml
+++ b/pom.xml
@@ -112,10 +112,17 @@
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-compiler-plugin</artifactId>
 				<configuration>
-					<!-- Spotless touches file timestamps during apply, which causes Maven's
-					     incremental compiler to recompile files without re-running Lombok's
+					<!-- Declare Lombok as an explicit annotationProcessorPath so it runs
+					     correctly on incremental recompiles (e.g. after Spotless touches
+					     timestamps). Without this, incremental compilation skips Lombok's
 					     annotation processor, breaking @Slf4j log field generation. -->
-					<useIncrementalCompilation>false</useIncrementalCompilation>
+					<annotationProcessorPaths>
+						<path>
+							<groupId>org.projectlombok</groupId>
+							<artifactId>lombok</artifactId>
+							<version>1.18.38</version>
+						</path>
+					</annotationProcessorPaths>
 				</configuration>
 			</plugin>
 			<plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -120,7 +120,7 @@
 						<path>
 							<groupId>org.projectlombok</groupId>
 							<artifactId>lombok</artifactId>
-							<version>1.18.38</version>
+							<version>${lombok.version}</version>
 						</path>
 					</annotationProcessorPaths>
 				</configuration>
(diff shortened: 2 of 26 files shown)


src/main/java/com/contrast/labs/ai/mcp/contrast/config/ContrastSDKFactory.java
dougj-contrast, line 104: there is an apache open source library that provides this url validation

https://commons.apache.org/proper/commons-validator/

https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/UrlValidator.html

```
<dependency>
    <groupId>commons-validator</groupId>
    <artifactId>commons-validator</artifactId>
    <version>1.7</version> <!-- Use the latest version -->
</dependency>

```

ChrisEdwards: Implemented in PR #100

dougj-contrast (approved):

PR #97, opened 2026-04-20
AIML-591: Add Checkstyle for automated style enforcement

ChrisEdwards:
## Summary
- Adds Checkstyle to the Maven `validate` phase with three rules enforced at `error` severity: no wildcard imports, no fully-qualified class names, no magic numbers
- Fixes all pre-existing violations across the codebase (constants extracted, FQCNs replaced with imports)
- Updates `make check` to run both Spotless and Checkstyle, and documents rules in CLAUDE.md with an explicit prohibition on relaxing them

## Why This Change Exists

Analysis of historical PR review comments found that ~70% were style/convention issues that could have been caught automatically. Reviewers were spending time flagging wildcard imports, fully-qualified class names used inline, and raw numeric literals — all of which are mechanical checks that belong in CI, not code review.

The goal is to move style enforcement from expensive, inconsistent human review to cheap, consistent automated checks. This frees reviewers to focus on logic, architecture, and correctness.

## Approach We Chose

Three Checkstyle rules were selected based on the most common PR comment themes:

1. **`AvoidStarImport`** — Wildcard imports were flagged repeatedly (e.g., PR #30). No current violations; the rule prevents regressions.
2. **`RegexpSinglelineJava`** — Fully-qualified class names in code were flagged in PRs #27 and #39. Fixed in `GetVulnerabilityTool` where `com.contrastsecurity.sdk.ContrastSDK` was used inline in two private method signatures.
3. **`MagicNumber`** — Raw numeric literals were flagged in PRs #38 and #39. Fixed by extracting constants throughout the codebase and centralizing shared values in `ValidationConstants`.

Checkstyle runs in the `validate` phase so it fires before compilation, giving immediate feedback. `make check` was updated to invoke `mvn validate` rather than just `spotless:check`, so both Spotless and Checkstyle run together.

## Outcome of This Change

- `make check` now enforces style automatically — no configuration required from developers
- All three acceptance criteria from AIML-591 are met
- Zero violations in the current codebase; new violations will fail the build at `validate` phase
- CLAUDE.md documents the rules and explicitly prohibits relaxing them without permission, so AI agents cannot work around failing checks

## Reviewer Guide

1. Start with `checkstyle.xml` — this is the single source of truth for all rules
2. Review `pom.xml` for plugin configuration and dependency pinning
3. Review `Makefile` to confirm `make check` now runs `mvn validate`
4. Scan the code changes — they are all mechanical (extract constant, replace FQCN with import)
[...]

PR #98, opened 2026-04-25
AIML-674: Harden integration tests before Hosted MCP Server

ChrisEdwards:
_Rebased onto main after #97 merged._

## Note
These are test updates only, no production code changes other than fixing a tool description about case-insensitive searches. Don't feel like you have to review this in great detail. Its a lot of changes, but they are tests only.

## Summary
- Codifies assertion-quality standards for integration tests in `CLAUDE.md` so future contributors (human or AI) cannot regress them silently.
- Hardens 13 `*IT.java` files end-to-end so each integration test verifies real SDK/API contract behavior — filter predicates, sort ordering, pagination boundaries, error-message shape, and field mapping — instead of confirming auth still works.
- Replaces silent-skip anti-patterns (`Assumptions.assumeTrue`, `if (data.isEmpty()) return;`) with fail-loudly preconditions, and replaces shape-only assertions (`isNotNull`) with content-level and behavioral assertions.
- Includes one small product-side fix: `GetSastProjectTool` documentation now correctly states that project-name matching is case-insensitive (uncovered while hardening `GetSastProjectToolIT`).

## Why This Change Exists

This branch is the test-quality prerequisite for the Hosted MCP Server epic (AIML-110). Before we expose this MCP server as a Contrast-hosted product, we need integration tests that catch real contract regressions — not tests that pass whenever credentials work and an HTTP 200 comes back.

Auditing the existing `*IT.java` suite revealed three classes of degraded tests:

1. **Silent-skip anti-pattern.** Many tests started with `assumeTrue(data != null && !data.isEmpty(), ...)` or `if (data.isEmpty()) return;`. When seeded test data drifts in the integration environment, these tests turn into no-ops and CI stays green. We had several tests effectively skipping in CI.
2. **Exercise ≠ verify.** Tests that called the tool, asserted `isNotNull()` on the response, and emitted a `log.info("✓ filter works")` line — but never asserted that the filter actually filtered, the sort actually sorted, or the pagination actually paginated. A `log.info` is not a test.
3. **Dual-path "always green" tests.** Patterns like `if (response.isSuccess()) { assertA } else { assertB }` treated both outcomes as a pass, so the test could not fail.

If the Hosted MCP Server ships against a test suite that doesn't actually verify the SDK/API contract, regressions land in production unseen. The cost of finding contract bugs in a hosted multi-tenant service is far higher than the cost of fixing them now.

## Approach We Chose

Two parallel workstreams:

### 1. Codify the standard

`CLAUDE.md` gains two new sections:
- **Assertion quality (applies to every test)** — eight rules covering deletion policy, mutation-check thinking, AAA completeness, fail-fast on missing data, behavioral over shape assertions, deterministic expectations, and "populated ≠ non-null".
- **Testing anti-patterns (do not reintroduce)** — eleven specific patterns that have been removed from this branch and must not return.
- **Integration Test Standards** — a dedicated subsection under "Testing Requirements Before Moving to Review" listing the required assertion patterns per test type (filter, sort, pagination, combined filters, error path, field mapping, populate claim, deterministic behavior), the data-dependency rule, and four canonical example IT classes to emulate.

The rationale for codifying this in `CLAUDE.md` rather than a separate doc: AI agents read `CLAUDE.md` on every session, so the standard is enforced at the point of contact. The anti-patterns list specifically calls out the failure modes we just fixed, so future agents don't reintroduce them.

### 2. Harden each IT systematically

One commit per IT (cleanly bisectable). Each commit follows the same playbook:

- Replace `Assumptions.assumeTrue(...)` and silent `return` with `assertThat(data).as("requires seeded X — see INTEGRATION_TESTS.md").isNotEmpty()` as the first assertion. Missing data now fails CI with an actionable message.
- Replace `assertThat(filtered).isNotEmpty()` with `assertThat(filtered).allSatisfy(item -> assertThat(item.field())...)` — proves every result matches the filter predicate, not just one.
- Replace manual sort-order loops with `assertThat(items).isSortedAccordingTo(Comparator.comparing(...))`.
- Replace `if (response.isSuccess()) ... else ...` dual-path tests by splitting them into two named tests, each with one expected outcome.
- For "_populate_*" / "_return_*" tests, replace `isNotNull()` on the claimed field with `isNotBlank()` for strings and `isNotEmpty()` for collections, plus content-level assertions.
- For error-path tests, assert the full validation message shape (e.g., `"projectName is required"`) and additionally assert `noneMatch("Contrast API error")` to ensure validation errors aren't masquerading as 5xx responses.
- For pagination tests, fetch page N+1 and verify it is disjoint from page N by ID, with item count = `min(pageSize, remaining)`.
- For field-mapping tests, verify response values reflect inputs (e.g., the response's `appID` equals the filter's `appId`).

Each commit message is suffixed with the originating bead (`mcp-wv00.N`) so the work is traceable back to the parent task tree.

## Outcome of This Change

- 13 IT files now genuinely verify SDK/API contract behavior. If we delete production logic under test, the corresponding IT goes red — the mutation-check property the new standard demands.
- Silent skips are gone. Missing seeded data fails CI loudly with a pointer to `INTEGRATION_TESTS.md`.
- The standard is documented and enforced through `CLAUDE.md`, so the next agent (human or AI) gets the same playbook on day one.
- Net diff: +4,780 / −1,431 across 16 files (13 IT files, `CLAUDE.md`, `TestDataDiscoveryHelper`, `GetSastProjectTool`).

## Reviewer Guide

The diff is large but mechanical and repetitive. Recommended order:

1. **`CLAUDE.md` first** — read the new "Assertion quality", "Testing anti-patterns", and "Integration Test Standards" sections. These define what every IT change is supposed to look like.
2. **One canonical IT** — pick `src/test/java/.../coverage/GetRouteCoverageToolIT.java` (cited in `CLAUDE.md` as the canonical pagination + filter + edge-cases example). Read it end-to-end. Once you understand the pattern, the other 12 IT files are variations on the same theme.
3. **Remaining IT files** — skim. Look for any place where `isNotNull` / `isNotEmpty` survives without a content-level companion assertion, or any `if (success) … else …` you can find. None should remain.
4. **`GetSastProjectTool.java`** — three-line product fix. Verify the description matches actual SDK behavior.
5. **`TestDataDiscoveryHelper.java`** — small refactor to prefer apps with vulnerable libraries. Confirm fallback behavior is preserved.

Per-commit review also works well — each commit is one IT and fully self-contained.

### 1. Problem framing — why ITs were drifting

The previous IT pattern was "call the tool, confirm no exception, confirm response is non-null, log a checkmark." This passes whenever:
- The user is authenticated and the server returns 200.
- The seeded data still exists (or the test silently skips).
- The response object is constructed (which always happens regardless of filter/sort correctness).

None of those conditions exercise the SDK/API contract. A regression where a filter parameter is silently dropped, a sort comparator is inverted, or a pagination cursor is off-by-one would not turn any of these tests red.

### 2. Codifying the new standard (`CLAUDE.md`)

`CLAUDE.md` is the file every AI agent reads first. By placing the rules and anti-patterns there — with concrete examples and named canonical files — the standard is enforced at the point of contact. The anti-patterns list is intentionally specific (e.g., "filter test that never calls `allSatisfy`") so future violations are recognizable in code review.

### 3. Per-IT hardening (13 commits)

Each commit replaces the same set of patterns:

| Old pattern | New pattern |
|---|---|
| `assumeTrue(data != null && !data.isEmpty(), ...)` | `assertThat(data).as("requires seeded X").isNotEmpty()` (first assertion) |
| `if (data.isEmpty()) return;` | Same as above |
| `assertThat(filtered).isNotEmpty(); log.info("✓ filter")` | `assertThat(filtered).allSatisfy(...)` |
| Manual sort-order `for` loop | `isSortedAccordingTo(Comparator.comparing(...))` |
| `if (success) {...} else {...}` | Split into two named tests, each deterministic |
| `assertThat(field).isNotNull()` (on a "_populate_" test) | `assertThat(field).isNotBlank()` / `isNotEmpty()` + content check |
| Pagination via `isNotEmpty()` only | Page N+1 disjoint from page N by ID, `min(pageSize, remaining)` |
| Error-path `errors().anyMatch(e -> e.contains("paramName"))` | Full message shape, plus `noneMatch("Contrast API error")` |

### 4. Supporting changes

**`GetSastProjectTool.java` (3-line fix)** — While hardening `GetSastProjectToolIT`, the test exercising case-insensitive lookups passed against a tool whose Javadoc claimed names were case-sensitive. The SDK behavior was correct; the documentation was wrong. Fixed the `@Tool` description and `@ToolParam` description to match real behavior.

**`TestDataDiscoveryHelper.java`** — Previously the helper returned the first app with any libraries. The SCA tests need an app with a *vulnerable* library to exercise CVE surfacing, so the helper now scans up to `maxAppsToCheck` apps preferring one with vulnerabilities, falling back to any-libraries-app only if no vulnerable-lib app is found within the limit. Logging is added on the fallback path.

### 5. Edge cases and safeguards

- **Validation messages** — error-path tests now assert the full message including the list of valid options where applicable, not just the parameter name. This prevents false positives where any error happens to contain the param name.
- **5xx vs validation** — error-path tests additionally assert `noneMatch("Contrast API error")` to ensure validation errors are never masquerading as the 5xx-mapped string.
- **Determinism** — every test asserts a single expected outcome. Where a behavior previously had two legitimate outcomes (e.g., success or graceful-not-found), the test was split into two named tests, each with one expected outcome.

### Standards
- `CLAUDE.md` — new "Assertion quality" rules, new "Testing anti-patterns" list, new "Integration Test Standards" section with required patterns, data-dependency rule, and canonical examples; minor strengthening of the existing test-naming bullet.

### Integration tests (13 files)
- `tool/application/GetSessionMetadataToolIT.java`
- `tool/application/SearchApplicationsToolIT.java`
- `tool/attack/GetProtectRulesToolIT.java`
- `tool/attack/SearchAttacksToolIT.java`
- `tool/coverage/GetRouteCoverageToolIT.java`
- `tool/library/ListApplicationLibrariesToolIT.java`
- `tool/library/ListApplicationsByCveToolIT.java`
- `tool/sast/GetSastProjectToolIT.java`
- `tool/sast/GetSastResultsToolIT.java`
- `tool/vulnerability/GetVulnerabilityToolIT.java`
- `tool/vulnerability/ListVulnerabilityTypesToolIT.java`
- `tool/vulnerability/SearchAppVulnerabilitiesToolIT.java`
- `tool/vulnerability/SearchVulnerabilitiesToolIT.java`

### Test infrastructure
- `util/TestDataDiscoveryHelper.java` — prefer apps with vulnerable libraries; deterministic fallback to any-libraries-app.

### Product
- `tool/sast/GetSastProjectTool.java` — corrects `@Tool` and `@ToolParam` descriptions to state that project-name matching is case-insensitive (the SDK was already case-insensitive; only the documentation was wrong).

## Testing

**Automated:**
- `make check-test` — passes (unit tests + style + checkstyle, 0 violations).
- `make verify` — passes against the integration environment with seeded data; every IT now asserts its precondition up front and exercises real contract behavior.

**Mutation-check (informal):** For each IT, asked the question "if I deleted the production logic under test, would this go red?" and strengthened assertions until the answer was yes. The same check applies to reviewers.

**Failure-mode validation:** Confirmed that removing seeded data causes a clear `requires seeded X — see INTEGRATION_TESTS.md` failure (instead of a silent green pass).

**Important gaps:** None known. Integration tests run against the same shared environment used by all `*IT.java` tests; new precondition assertions surface seed drift loudly rather than skipping.

- **Risk: tighter preconditions may surface seed drift in CI.** This is intentional — the previous behavior was to pass anyway. If a precondition fails, the message points to `INTEGRATION_TESTS.md` for the seed contract. This is a feature, not a regression.
- **No runtime risk.** The only product-side change is a documentation correction in `GetSastProjectTool` (`@Tool` / `@ToolParam` descriptions). Behavior is unchanged; the SDK was already case-insensitive.
- **No rollout concerns.** This branch ships only to the test suite and developer documentation.
- **Follow-up:** AIML-110 (Hosted MCP Server) can now build on a test suite that genuinely catches contract regressions. Subsequent work should adopt the same patterns from day one — the canonical examples in `CLAUDE.md` provide the template.

### DIFF
--- a/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/sast/GetSastProjectTool.java
+++ b/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/sast/GetSastProjectTool.java
@@ -51,13 +51,14 @@ public class GetSastProjectTool extends SingleTool<GetSastProjectParams, ScanPro
           Usage examples:
           - Get project: projectName="my-application"
 
-          Note: Project names are case-sensitive and must match exactly.
+          Note: Project name matching is case-insensitive. The returned project's name field
+          reflects the canonical casing as stored in Contrast.
 
           Related tools:
           - get_scan_results: Get SARIF results for a project's latest scan
           """)
   public SingleToolResponse<ScanProject> getScanProject(
-      @ToolParam(description = "Scan project name (case-sensitive, must match exactly)")
+      @ToolParam(description = "Scan project name (matched case-insensitively)")
           String projectName) {
     return executePipeline(() -> GetSastProjectParams.of(projectName));
   }
--- a/src/test/java/com/contrast/labs/ai/mcp/contrast/util/TestDataDiscoveryHelper.java
+++ b/src/test/java/com/contrast/labs/ai/mcp/contrast/util/TestDataDiscoveryHelper.java
@@ -88,6 +88,11 @@ public class TestDataDiscoveryHelper {
     int appsChecked = 0;
     int actualMaxToCheck = Math.min(applications.size(), maxAppsToCheck);
 
+    // Prefer an app whose libraries include a populated vulnerabilities array so SCA tests can
+    // exercise severity breakdowns and CVE surfacing. Fall back to the first app with any
+    // libraries if no vulnerable-lib app exists within maxAppsToCheck.
+    ApplicationWithLibraries firstAnyLibMatch = null;
+
     for (Application app : applications) {
       if (appsChecked >= actualMaxToCheck) {
         logger.info("Reached max apps to check ({}), stopping search", actualMaxToCheck);
@@ -104,36 +109,52 @@ public class TestDataDiscoveryHelper {
 
       try {
         var libraries = IntegrationTestDataCache.getLibraries(orgId, app.getAppId(), sdkExtension);
-        if (libraries != null && !libraries.isEmpty()) {
+        if (libraries == null || libraries.isEmpty()) {
+          continue;
+        }
+
+        boolean hasVulnerableLibrary = false;
+        String vulnerableCveId = null;
+        for (LibraryExtended lib : libraries) {
+          if (lib.getVulnerabilities() != null && !lib.getVulnerabilities().isEmpty()) {
+            hasVulnerableLibrary = true;
+            var firstVuln = lib.getVulnerabilities().get(0);
+            if (firstVuln.getName() != null && firstVuln.getName().startsWith("CVE-")) {
+              vulnerableCveId = firstVuln.getName();
+            }
+            break;
+          }
+        }
+
+        if (hasVulnerableLibrary) {
           logger.info(
-              "✓ Found application with {} library/libraries: {} (ID: {})",
+              "✓ Found application with {} library/libraries (vulnerable, CVE={}): {} (ID: {})",
               libraries.size(),
+              vulnerableCveId,
               app.getName(),
               app.getAppId());
+          return Optional.of(new ApplicationWithLibraries(app, libraries, true, vulnerableCveId));
+        }
 
-          // Check for vulnerable libraries
-          boolean hasVulnerableLibrary = false;
-          String vulnerableCveId = null;
-
-          for (LibraryExtended lib : libraries) {
-            if (lib.getVulnerabilities() != null && !lib.getVulnerabilities().isEmpty()) {
-              hasVulnerableLibrary = true;
-              var firstVuln = lib.getVulnerabilities().get(0);
-              if (firstVuln.getName() != null && firstVuln.getName().startsWith("CVE-")) {
-                vulnerableCveId = firstVuln.getName();
-                logger.info("  ✓ Has vulnerable library with CVE: {}", vulnerableCveId);
-                break;
-              }
-            }
-          }
-
-          return Optional.of(
-              new ApplicationWithLibraries(app, libraries, hasVulnerableLibrary, vulnerableCveId));
+        if (firstAnyLibMatch == null) {
+          logger.debug(
+              "Remembering app as non-vulnerable fallback: {} (ID: {})",
+              app.getName(),
+              app.getAppId());
+          firstAnyLibMatch = new ApplicationWithLibraries(app, libraries, false, null);
         }
       } catch (IOException e) {
         logger.warn("Error checking libraries for app {}: {}", app.getAppId(), e.getMessage());
         // Continue to next app
       }
+    }
+
+    if (firstAnyLibMatch != null) {
+      logger.info(
+          "✓ Falling back to non-vulnerable app with libraries: {} (ID: {})",
+          firstAnyLibMatch.getApplication().getName(),
+          firstAnyLibMatch.getApplication().getAppId());
+      return Optional.of(firstAnyLibMatch);
     }
 
     logger.warn("No application with libraries found after checking {} apps", appsChecked);
--- a/src/test/java/com/contrast/labs/ai/mcp/contrast/util/TestDataDiscoveryHelper.java
+++ b/src/test/java/com/contrast/labs/ai/mcp/contrast/util/TestDataDiscoveryHelper.java
@@ -88,9 +88,10 @@ public class TestDataDiscoveryHelper {
     int appsChecked = 0;
     int actualMaxToCheck = Math.min(applications.size(), maxAppsToCheck);
 
-    // Prefer an app whose libraries include a populated vulnerabilities array so SCA tests can
-    // exercise severity breakdowns and CVE surfacing. Fall back to the first app with any
-    // libraries if no vulnerable-lib app exists within maxAppsToCheck.
+    // Prefer an app whose libraries include a populated vulnerabilities array AND surface a
+    // CVE-prefixed vulnerability name so SCA tests can exercise severity breakdowns and CVE
+    // lookups. Fall back tiers: vulnerable-but-no-CVE-name, then any app with libraries.
+    ApplicationWithLibraries firstVulnerableNoCveMatch = null;
     ApplicationWithLibraries firstAnyLibMatch = null;
 
     for (Application app : applications) {
@@ -116,17 +117,22 @@ public class TestDataDiscoveryHelper {
         boolean hasVulnerableLibrary = false;
         String vulnerableCveId = null;
         for (LibraryExtended lib : libraries) {
-          if (lib.getVulnerabilities() != null && !lib.getVulnerabilities().isEmpty()) {
-            hasVulnerableLibrary = true;
-            var firstVuln = lib.getVulnerabilities().get(0);
-            if (firstVuln.getName() != null && firstVuln.getName().startsWith("CVE-")) {
-              vulnerableCveId = firstVuln.getName();
+          if (lib.getVulnerabilities() == null || lib.getVulnerabilities().isEmpty()) {
+            continue;
+          }
+          hasVulnerableLibrary = true;
+          for (var vuln : lib.getVulnerabilities()) {
+            if (vuln.getName() != null && vuln.getName().startsWith("CVE-")) {
+              vulnerableCveId = vuln.getName();
+              break;
             }
+          }
+          if (vulnerableCveId != null) {
             break;
           }
         }
 
-        if (hasVulnerableLibrary) {
+        if (hasVulnerableLibrary && vulnerableCveId != null) {
           logger.info(
               "✓ Found application with {} library/libraries (vulnerable, CVE={}): {} (ID: {})",
               libraries.size(),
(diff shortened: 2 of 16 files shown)


src/main/java/com/contrast/labs/ai/mcp/contrast/tool/sast/GetSastProjectTool.java
```diff
@@ -51,13 +51,14 @@ public class GetSastProjectTool extends SingleTool<GetSastProjectParams, ScanPro
...
-          Note: Project names are case-sensitive and must match exactly.
+          Note: Project name matching is case-insensitive. The returned project's name field
+          reflects the canonical casing as stored in Contrast.
 
           Related tools:
           - get_scan_results: Get SARIF results for a project's latest scan
           """)
   public SingleToolResponse<ScanProject> getScanProject(
-      @ToolParam(description = "Scan project name (case-sensitive, must match exactly)")
+      @ToolParam(description = "Scan project name (matched case-insensitively)")
```
ChrisEdwards, line 61: FYI: The new test assertions found that the scan project name was case insensitive, not case-sensitive.

ChrisEdwards (commented):
Starting review...

ChrisEdwards (commented):
## 📋 Review Summary

This PR successfully hardens integration tests by implementing robust assertion standards and replacing silent skips with fail-loudly preconditions. The changes significantly improve the mutation-check property of the test suite and provide clear guidance for future contributions in CLAUDE.md.

## 🔍 General Feedback

- **Standards**: Excellent work codifying the assertion standards in CLAUDE.md. This is a high-signal way to ensure quality for both human and AI contributors.
- **Discovery**: The discovery pattern in the integration tests is a very robust way to handle environment-dependent data while still failing fast if the environment is misconfigured.
- **Documentation**: The fix in GetSastProjectTool documentation is a great catch and matches real SDK behavior.

src/test/java/com/contrast/labs/ai/mcp/contrast/util/TestDataDiscoveryHelper.java
```diff
@@ -104,38 +109,54 @@ public static Optional<ApplicationWithLibraries> findApplicationWithLibraries(
 
       try {
         var libraries = IntegrationTestDataCache.getLibraries(orgId, app.getAppId(), sdkExtension);
-        if (libraries != null && !libraries.isEmpty()) {
+        if (libraries == null || libraries.isEmpty()) {
+          continue;
+        }
+
```
ChrisEdwards, line 116: [LOW] The current logic only checks the first vulnerability of a library for a CVE ID. If the first vulnerability has a non-CVE name but a subsequent one has a CVE ID, the `vulnerableCveId` will remain null. Consider looping through the vulnerabilities to find the first one starting with "CVE-".

```suggestion
        for (LibraryExtended lib : libraries) {
          if (lib.getVulnerabilities() != null && !lib.getVulnerabilities().isEmpty()) {
            hasVulnerableLibrary = true;
            for (var vuln : lib.getVulnerabilities()) {
              if (vuln.getName() != null && vuln.getName().startsWith("CVE-")) {
                vulnerableCveId = vuln.getName();
                break;
              }
            }
            break;
          }
        }
```

ChrisEdwards (commented):
## 📋 Review Summary

This PR successfully hardens integration tests by implementing robust assertion standards and replacing silent skips with fail-loudly preconditions. The changes significantly improve the mutation-check property of the test suite and provide clear guidance for future contributions in CLAUDE.md.

## 🔍 General Feedback

- **Standards**: Excellent work codifying the assertion standards in CLAUDE.md. This is a high-signal way to ensure quality for both human and AI contributors.
- **Discovery**: The discovery pattern in the integration tests is a very robust way to handle environment-dependent data while still failing fast if the environment is misconfigured.
- **Documentation**: The fix in GetSastProjectTool documentation is a great catch and matches real SDK behavior.

src/test/java/com/contrast/labs/ai/mcp/contrast/util/TestDataDiscoveryHelper.java
ChrisEdwards: Done in af599f6 — extended beyond the suggestion: now scans every vulnerable library (not just the first) and every vuln within each library to find a `CVE-`-prefixed name. Added a tiered fallback: vulnerable+CVE > vulnerable-no-CVE > any-libs. The 11 `ListApplicationsByCveToolIT` tests pass against the live API.

Alex-Contrast:
I'm approving, but thought it would be good to call out that a certain org / env is required for integration tests to pass that isn't described in INTEGRATION_TESTS.md.

Alex-Contrast (approved):

ChrisEdwards:
> I'm approving, but thought it would be good to call out that a certain org / env is required for integration tests to pass that isn't described in INTEGRATION_TESTS.md.

It doesn't have to be a specific org. It just needs to be one that has apps with scenarios matching the testing requirements. It doesn't hardcode specific applications, it discovers them. If a scenario is missing, it fails loudly. (It used to silently skip).

PR #102, opened 2026-04-28
AIML-693: Expand checkstyle rules from 3 to 18 to mechanically enforce coding standards

ChrisEdwards:
## Summary
- Expands `checkstyle.xml` from 3 rules to 18, so the build mechanically enforces coding standards that CLAUDE.md previously relied on agent discipline to follow
- Migrates 5 files from manual `Logger` field declarations to Lombok `@Slf4j`
- Rewrites 16 `mock(X.class)` sites to use Mockito's no-arg `mock()` with explicit-type LHS

## Why This Change Exists
`checkstyle.xml` previously enforced only three rules (`AvoidStarImport`, `RegexpSinglelineJava` for FQCNs, `MagicNumber`). Every other standard in CLAUDE.md — use AssertJ not JUnit assertions, use `.toList()` not `Collectors.toList()`, use `@Slf4j` not manual Logger fields, use `mock()` not `mock(X.class)`, etc. — depended entirely on the agent reading and following prose. That's a weak guarantee: agents drift, new contributors miss it, and nothing in the build catches regressions.

Mechanical enforcement closes that gap. The rule is either enforced or it isn't; there's no "I forgot to check CLAUDE.md" failure mode.

## Approach We Chose
One commit per rule group, each commit adds the rule(s) to `checkstyle.xml` and fixes any pre-existing violations in the same change. This keeps each commit reviewable in isolation and makes it easy to bisect if a rule causes unexpected pain.

**Rules added in commit order:**

1. **Zero-violation regex rules** (4) — `Collectors.toList()`, `.size()` vs 0 comparisons, JUnit assertions, `Assumptions.assume`. All had zero existing violations; adding them prevents future drift.
2. **`mock(X.class)` regex rule + 16 site rewrites** — Mockito's no-arg `mock()` infers from the LHS type. `var x = mock(X.class)` → `Foo x = mock();` (13 sites); field-init `mock(X.class)` → `mock()` (3 sites in `Anonymous*Builder`).
3. **Manual `Logger` regex rule + 5 `@Slf4j` conversions** — Lombok `@Slf4j` generates an identical `private static final Logger log` field. Converted `McpContrastApplication` and 4 test-utility classes.
4. **13 built-in Checkstyle modules** — `UnusedImports`, `RedundantImport`, `EmptyCatchBlock`, `MissingOverride`, `EqualsHashCode`, `StringLiteralEquality`, `FallThrough`, `DefaultComesLast`, `MissingSwitchDefault`, `UpperEll`, `SimplifyBooleanExpression`, `SimplifyBooleanReturn`, `ModifiedControlVariable`. The codebase had zero violations against all 13; they're guard rails only.
5. **CLAUDE.md update** — Expands the Checkstyle bullet from the 3-rule list to an 18-rule grouped summary. Other coding-standards bullets unchanged (checkstyle is the safety net; CLAUDE.md remains preventative guidance).

All 6 regex rules use `ignoreComments="true"` so explanatory comments referencing a banned pattern don't trigger false positives.

## Outcome of This Change
- `make check` now rejects the most common coding-standard regressions at build time, not review time
- Agents (and humans) get an immediate, actionable error message at the violation site rather than a vague "check CLAUDE.md" reminder
- Zero net behavior change at runtime — all edits are mechanical refactors with identical semantics (`@Slf4j` → same Logger field, `mock()` → same Mockito stub)
- No metrics gathered beyond the pre-existing violation counts (all were ≤16; none exceeded the 20-violation threshold that would have deferred a rule to a follow-up)

## Reviewer Guide
1. **Start with `checkstyle.xml`** — skim the 5 new rule groups. Verify each module has `severity="error"` and the regex rules have `ignoreComments="true"`.
2. **Skim the `mock()` rewrites** — each changed line should be `Foo x = mock();` (explicit type, no `.class`). The key invariant: the type on the LHS must match what was previously inside `mock(…)`.
3. **Skim the `@Slf4j` conversions** — each converted file should have `@Slf4j` on the class, no `Logger` field, no `LoggerFactory` import, and all `logger.` calls renamed to `log.`.
4. **`CLAUDE.md`** — verify the Checkstyle bullet now describes the full rule set and is grouped by purpose. Everything else in the file is unchanged.

### 1. Problem framing
CLAUDE.md documented ~18 coding standards but `checkstyle.xml` enforced only 3. The gap meant regressions slipped through until review.

### 2. Core implementation
Six `RegexpSinglelineJava` modules and 13 built-in modules added to the `<TreeWalker>` block in `checkstyle.xml`. All at `error` severity so violations fail `make check` (which runs in the Maven `validate` phase, before any compilation).

### 3. Supporting changes
Pre-existing violations fixed alongside each rule: 16 `mock(X.class)` sites, 5 manual `Logger` declarations. The built-in rules had zero pre-existing violations.

### 4. Edge cases and safeguards
- `mock(Rules.Rule.class)` (inner-class pattern with a dot) does **not** match `mock\([A-Z]\w+\.class\)` — the dot breaks the pattern. This is correct: inner-class mocks are rare and the rule targets the common simple-name form.
- All regex patterns include `\b` word boundaries where needed to prevent false positives on class names that end in a banned suffix (e.g. `MyAssumptions`).
- `ignoreComments="true"` on every regex rule prevents comment-only references to banned patterns from triggering errors.

### 5. How the tests prove it
`make check-test` passes (689 unit tests, 0 checkstyle violations). The `@Slf4j` and `mock()` rewrites are semantically identical to what they replaced — no behavior tests required.

### `checkstyle.xml`
- Added 6 `RegexpSinglelineJava` modules and 13 built-in modules, all at `error` severity inside `<TreeWalker>`

### `src/main/java/.../McpContrastApplication.java`
- Converted manual `Logger logger` field to `@Slf4j`; renamed `logger.` → `log.`

### `src/test/java/.../Anonymous*Builder.java` (3 files)
- Field-init `mock(X.class)` → `mock()` (type inferred from field declaration)

### `src/test/java/.../` (test utility classes, 4 files)
- `IntegrationTestDataCache`, `TestDataDiscoveryHelper`, `IntegrationTestDiskCache`, `IntegrationTestConfig` — converted manual `Logger` to `@Slf4j`

### `src/test/java/.../` (tool test files, 4 files)
- `ListApplicationsByCveToolTest`, `GetSessionMetadataToolTest`, `GetVulnerabilityToolTest`, `ListVulnerabilityTypesToolTest` — `var x = mock(X.class)` → `X x = mock()`

### `CLAUDE.md`
- Checkstyle bullet expanded from 3-rule list to 18-rule grouped summary

## Testing
- `make check`: 0 checkstyle violations
- `make check-test`: 689 unit tests pass
- Integration tests not run locally (require Contrast credentials); CI covers them

- **Risk**: if a future contributor introduces a pattern that matches a new regex rule, they'll get an immediate build failure — this is the intended behavior, not a risk
- **Follow-up**: AST-level test-shape rules (e.g. "filter test missing `allSatisfy`") are out of scope here; they'd require a custom Checkstyle check or ArchUnit rule
- **Follow-up**: `mock(Rules.Rule.class)` (inner-class form) is not caught by the current regex; a more complex pattern could be added if this becomes a recurring issue

Refs: [AIML-693](https://contrast.atlassian.net/browse/AIML-693)

[AIML-693]: https://contrast.atlassian.net/browse/AIML-693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/checkstyle-suppressions.xml
+++ b/checkstyle-suppressions.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0"?>
+<!DOCTYPE suppressions PUBLIC
+  "-//Checkstyle//DTD SuppressionFilter Configuration 1.2//EN"
+  "https://checkstyle.org/dtds/suppressions_1_2.dtd">
+<suppressions>
+  <suppress checks="MagicNumber" files="src[\\/]test[\\/]"/>
+</suppressions>
--- a/pom.xml
+++ b/pom.xml
@@ -252,9 +252,11 @@
 				</dependencies>
 				<configuration>
 					<configLocation>checkstyle.xml</configLocation>
+					<suppressionsLocation>checkstyle-suppressions.xml</suppressionsLocation>
 					<consoleOutput>true</consoleOutput>
 					<failsOnError>true</failsOnError>
 					<linkXRef>false</linkXRef>
+					<includeTestSourceDirectory>true</includeTestSourceDirectory>
 				</configuration>
 				<executions>
 					<execution>
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -183,10 +183,12 @@ When creating or modifying MCP tools:
 - No fully-qualified class names - use imports
 - `isEmpty()` not `size() > 0` for collections
 
-**Checkstyle:** Three rules enforced at `error` severity (run in `validate` phase via `make check`):
-- `AvoidStarImport` — no wildcard imports
-- `RegexpSinglelineJava` — no fully-qualified class names in code; use imports
-- `MagicNumber` — no raw numeric literals; use named constants (HTTP status codes and -1/0/1/2/100 are ignored). **Before writing any numeric literal**, check `ValidationConstants` first — it has `DEFAULT_PAGE_SIZE`, `MAX_PAGE_SIZE`, `API_MAX_PAGE_SIZE`, `DEFAULT_LIBRARY_OBS_PAGE_SIZE`, `MIN_PAGE`, `DEFAULT_PAGE`. If no existing constant fits, declare `private static final int MY_CONSTANT = <value>` in the same class.
+**Checkstyle:** 18 rules enforced at `error` severity (run in `validate` phase via `make check`). The full list lives in `checkstyle.xml`. Highlights:
+- **Imports:** `AvoidStarImport`, `UnusedImports`, `RedundantImport`, `RegexpSinglelineJava` (no FQCN — use imports)
+- **Numbers:** `MagicNumber` — no raw numeric literals; use named constants (HTTP status codes and -1/0/1/2/100 are ignored). **Before writing any numeric literal**, check `ValidationConstants` first — it has `DEFAULT_PAGE_SIZE`, `MAX_PAGE_SIZE`, `API_MAX_PAGE_SIZE`, `DEFAULT_LIBRARY_OBS_PAGE_SIZE`, `MIN_PAGE`, `DEFAULT_PAGE`. If no existing constant fits, declare `private static final int MY_CONSTANT = <value>` in the same class. `UpperEll` (`1L` not `1l`).
+- **Correctness:** `EmptyCatchBlock`, `MissingOverride`, `EqualsHashCode`, `StringLiteralEquality` (`s == "FOO"` is always wrong), `FallThrough`, `DefaultComesLast`, `MissingSwitchDefault`, `ModifiedControlVariable`
+- **Style:** `SimplifyBooleanExpression`, `SimplifyBooleanReturn`
+- **Codebase conventions (regex):** ban `Collectors.toList()` (use `.toList()`), `mock(X.class)` (use `mock()` with explicit-type LHS), `.size() > 0` (use `isEmpty()`), JUnit assertions in tests (use AssertJ), `Assumptions.assume*` (fail loudly), manual `Logger` fields (use `@Slf4j`)
 
 > ⛔ **PROHIBITED:** Modifying checkstyle rules, Spotless config, or any other linter/constraint config is **expressly forbidden** without explicit user permission. When code fails a check, fix the code — never relax the rule.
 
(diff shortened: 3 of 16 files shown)


ChrisEdwards:
I added this as my last PR had some issues that Doug found that should have been enforced.

ChrisEdwards:
Codex reviewed, and issues it found were fixed.

pom.xml
```diff
@@ -252,9 +252,11 @@
 				</dependencies>
 				<configuration>
 					<configLocation>checkstyle.xml</configLocation>
+					<suppressionsLocation>checkstyle-suppressions.xml</suppressionsLocation>
 					<consoleOutput>true</consoleOutput>
 					<failsOnError>true</failsOnError>
 					<linkXRef>false</linkXRef>
+					<includeTestSourceDirectory>true</includeTestSourceDirectory>
```
ChrisEdwards, line 259: FYI: Turned it on for tests.

checkstyle-suppressions.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<!DOCTYPE suppressions PUBLIC
+  "-//Checkstyle//DTD SuppressionFilter Configuration 1.2//EN"
+  "https://checkstyle.org/dtds/suppressions_1_2.dtd">
+<suppressions>
```
ChrisEdwards, line 5: FYI: Use of magic numbers don't make sense to enforce in tests. Assertions often assert counts and such.

JacobMagesHaskinsContrast (approved):
AI had these 2 suggestions about CLAUDE.md:

  1. Document (or suppress) the mock(X.class) method-argument pattern in CLAUDE.md — when mock(Foo.class) appears as a method argument (not an assignment target), no-arg mock() isn't a valid replacement, but the checkstyle error message says
  it is.
  2. Clarify suppression policy in CLAUDE.md — explicitly state whether adding <suppress> entries requires the same "explicit user permission" as modifying rules themselves.

Alex-Contrast (approved):

ChrisEdwards:
**Re: suggestion #1 (mock(X.class) arg-position)**

Good catch — the error message was only guiding toward `Foo x = mock();` which doesn't compile at argument position. Updated in 41721f2:

- `checkstyle.xml` error message now explicitly says: for argument positions, extract to a typed local first (`Foo x = mock(); when(x.method())...`)
- CLAUDE.md updated in both the Checkstyle bullet and the Testing section with the same guidance

ChrisEdwards:
**Re: suggestion #2 (suppression policy)**

Updated in 74722ee — the PROHIBITED block in CLAUDE.md now explicitly calls out `checkstyle-suppressions.xml`: a suppression relaxes a rule at a specific site, which is equally forbidden without explicit user permission.

PR #140, opened 2026-06-01
AIML-822 Publish core via JFrog OIDC release flow

ChrisEdwards:
**Jira:** [AIML-822](https://contrast.atlassian.net/browse/AIML-822)

## Summary
NOTE: Previously, I had implemented a complex workflow to publish to internal artifactory. This is a simplified workflow to publish to our public artifactory. A lot of the file changes (of the 17) are deletes of the pr walkthrough files we don't need. 

NOTE: There is a smoke test flow in the release workflow that is to test the artifactory credentials. I can't test them until they are merged because the security settings only allow it to run from main. I have ran all the tests I can to verify it, except the actual workflow. That will be don after merge.

---
Migrates `contrast-mcp-core` publication from the internal Artifactory workflow (secret-based auth, separate workflow file) to JFrog OIDC authentication through the existing `gradle-release.yml` workflow, and adds an `artifactory-smoke` mode for validating the OIDC pipeline without publishing a real release.

- Core artifact now publishes to public Artifactory via short-lived OIDC tokens instead of long-lived secrets
- Standalone `publish-core-artifactory.yml` workflow is removed in favor of a unified release workflow with a `mode` input
- Old internal Artifactory names (`contrastInternalRelease`, `contrastArtifactoryReleaseUrl`, etc.) are replaced throughout

## Why This Change Exists

The previous publish path used long-lived Artifactory credentials stored as GitHub secrets, split across a separate workflow file with a two-job verify-then-publish design. This was more complex than necessary and relied on static credentials that require manual rotation. OIDC federation eliminates stored credentials entirely and lets the release workflow handle both the GitHub Release and the Maven publication in a single job.

## Approach

Instead of adding OIDC support to the existing standalone workflow, the publish steps were folded into `gradle-release.yml` as part of the release job. A new `artifactory-smoke` mode was added so the OIDC pipeline can be validated end-to-end (connectivity, permissions, Gradle publication task) without cutting a real release. The smoke mode optionally publishes a `0.0.0-artifactory-smoke-*` coordinate for full round-trip verification.

The Gradle build config was updated to target a single `contrastPublicRelease` repository with a `contrastPublicMavenReleaseUrl` property, removing the internal/snapshot split. A SNAPSHOT guard now rejects publish attempts for non-release versions.

### `.github/workflows/gradle-release.yml`
- Added `mode` (release / artifactory-smoke) and `publish_smoke_artifact` workflow inputs
- New `artifactory-smoke` job validates OIDC connectivity, optionally publishes a smoke artifact
- Release job now includes JFrog CLI setup via OIDC and publishes `contrast-mcp-core` after building the release artifact and before creating the GitHub Release
- Release job gated with `if: inputs.mode == 'release'`

### `.github/workflows/publish-core-artifactory.yml`
- Deleted. The standalone two-job workflow is superseded by the unified release workflow.

### `contrast-mcp-core/build.gradle`
- Renamed repository from `contrastInternalRelease`/`contrastInternalSnapshot` to `contrastPublicRelease`
- Replaced `contrastArtifactoryReleaseUrl`/`contrastArtifactorySnapshotUrl` with single `contrastPublicMavenReleaseUrl` property
- Added explicit SNAPSHOT rejection guard before publish

### `hack/verify-public-workflow-alignment.sh`
- Assertions updated to validate the new JFrog OIDC flow (pinned action SHA, OIDC provider variable, public release URL variable, OIDC-derived credentials)
- Added `assert_line_order` helper to verify publish step ordering within the workflow
- Replaced old `publish-core-artifactory.yml` assertions with `assert_missing` check

### `docs/pr-walkthrough/`
- Removed 12 stale PR walkthrough HTML files (housekeeping, unrelated to the publish migration)

## Testing

- `make check-test` passes locally (includes `workflow-check` alignment gate with updated assertions)
- Smoke mode is designed for manual validation via `workflow_dispatch` in GitHub Actions
- No unit test changes needed as this is purely CI/CD workflow and Gradle config

[AIML-822]: https://contrast.atlassian.net/browse/AIML-822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hack/verify-public-workflow-alignment.sh
```diff
@@ -72,6 +72,23 @@ assert_missing() {
   log "status=pass assertion=\"${label}\" path=${path}"
```
ChrisEdwards, line 72: This is a temporary validation script that will go away very soon. No need to spend time on it.

contrast-mcp-core/build.gradle
```diff
@@ -22,14 +22,10 @@ dependencies {
...
 
-def artifactoryRepositoryName = version.toString().endsWith('-SNAPSHOT')
-    ? 'contrastInternalSnapshot'
-    : 'contrastInternalRelease'
-def artifactoryUrlPropertyName = version.toString().endsWith('-SNAPSHOT')
-    ? 'contrastArtifactorySnapshotUrl'
-    : 'contrastArtifactoryReleaseUrl'
-def artifactoryUrlFallback = 'https://artifactory-url-required.invalid'
-def configuredArtifactoryUrl = providers.gradleProperty(artifactoryUrlPropertyName)
+def publicMavenReleaseRepositoryName = 'contrastPublicRelease'
```
ChrisEdwards, line 25: Migrating this to work only on public artifactory and releases, no snapshots.

.github/workflows/gradle-release.yml
```diff
@@ -11,7 +25,74 @@ permissions:
   contents: read
 
 jobs:
+  artifactory-smoke:
```
ChrisEdwards, line 28: This is the job to smoke test atrifactory credentials...since this OIDC flow for artifactory is new (first time at contrast)

.github/workflows/gradle-release.yml
```diff
@@ -11,7 +25,74 @@ permissions:
...
+        env:
+          JF_URL: ${{ vars.CONTRAST_JFROG_URL }}
+        with:
+          oidc-provider-name: ${{ vars.JFROG_OIDC_PROVIDER_NAME }}
+
+      - name: Ping public Artifactory with OIDC credentials
+        run: jf rt ping
+
+      - name: Publish public Artifactory smoke artifact
+        if: ${{ inputs.publish_smoke_artifact }}
```
ChrisEdwards, line 78: Only try to publish a fake artifact to test that it works. (because it only allows publishing from main)

.github/workflows/gradle-release.yml
```diff
@@ -11,7 +25,74 @@ permissions:
...
+          smoke_version="0.0.0-artifactory-smoke-${GITHUB_RUN_ID}.${GITHUB_RUN_ATTEMPT}"
+          echo "Publishing smoke coordinate com.contrast.labs.ai.mcp:contrast-mcp-core:${smoke_version}"
+          ./gradlew \
+            :contrast-mcp-core:verifyCorePublicationMetadata \
+            :contrast-mcp-core:publishMavenJavaPublicationToContrastPublicReleaseRepository \
+            -Pversion="${smoke_version}" \
+            --no-daemon
+
   release:
+    if: ${{ inputs.mode == 'release' }}
```
ChrisEdwards, line 95: This is the actual release flow here. Publish to public artifactory.

.github/workflows/gradle-release.yml
```diff
@@ -93,6 +174,37 @@ jobs:
       - name: Build Release Artifact
         run: ./gradlew clean :contrast-mcp-stdio-app:bootJar -x test --no-daemon
 
+      - name: Validate public Artifactory release configuration
```
Alex-Contrast, line 177: Unclear if we want the attest provenance step here. I assume no, since you didn't add it, but thought it was worth a callout in case. 

> Will the hosted MCP server build will verify provenance on its contrast-mcp-core dependency?

Alex-Contrast (approved):

seschis (approved):
seems fine to me

.github/workflows/gradle-release.yml
ChrisEdwards: If you mean on the consuming side, yes, I will add that, but I hadn't yet since we hadn't been able to publish yet.

ChrisEdwards: Ok, I see what you mean. There are two attestations we are using. One is for the internal artifact for this PR...that uses pgp signing with an .asc file in artifactory. We chose that approach because it is idiomatic for gradle to pull and verify the file that way.

For the mcp-contrast artifact that customers can download and run, we only release that in github a a github release. We don't send it to maven, so we chose to keep it simple and use github attestation there. Its easier for the user. If we used this same pgp approach, we would need to publish the public key so users could use it, and give specific instructions how to verify the mcp-contrast jar. Its easier for the user to just use the gh cli to do it.

PR #143, opened 2026-06-02
[AIML-839] Migrate release versioning to Axion tag-derived model

ChrisEdwards:
**Jira:** [AIML-839](https://contrast.atlassian.net/browse/AIML-839)

## Summary
Migrate from the commit-based release versioning model to the Axion release plugin so the release workflow no longer pushes version commits to the protected `main` branch.

- Release versions now derive from immutable `vX.Y.Z` Git tags instead of `gradle.properties`
- Removes the `artifactory-smoke` mode and `setVersion` task
- Preserves all existing security controls (attestation, PGP signing, JFrog OIDC, Docker Content Trust)

## Why This Change Exists
GitHub repository rules on `main` reject direct pushes (GH013). The previous release workflow committed release-version and next-snapshot changes directly to `main`, which broke under these rules.

## Approach
Adopt the Axion release plugin, which derives the project version from Git tags at build time. The workflow creates a `vX.Y.Z` tag via Axion, checks it out, verifies the Gradle version matches, then builds and publishes from the tag. No commits are pushed to `main` during the release process.

An optional `release_version` workflow input allows explicit major/minor releases. Blank input auto-increments the patch version from the latest tag. The workflow handles rerun/failure recovery by reusing an existing tag if it already points at the current HEAD.

### Gradle build (`build.gradle`, `gradle.properties`)
- Added the Axion release plugin (`pl.allegro.tech.build.axion-release` v1.21.1) with `v` prefix and no separator
- Removed the `setVersion` task and `version=2.0.0-SNAPSHOT` from `gradle.properties`
- `printVersion` now prints Axion's computed version

### Release workflow (`.github/workflows/gradle-release.yml`)
- Replaced `mode` and `publish_smoke_artifact` inputs with a single optional `release_version` string input
- Removed the entire `artifactory-smoke` job
- Added release version input validation, tracked-worktree cleanliness checks, and a tag/version consistency gate
- Replaced the commit-tag-commit-push sequence with Axion tag creation and detached-HEAD checkout
- Moved GitHub Release creation to after Docker publishing (so it only appears once all artifacts succeed)

### Release docs (`RELEASING.md`)
- Updated to describe tag-based flow, operator instructions for blank vs explicit versions, manual release steps, and failure recovery guidance
- Removed references to `gradle.properties` version management and snapshot commits

## Testing
- Local `./gradlew -q printVersion` confirms Axion-derived snapshot version on untagged commits
- `make check-test` passes
- Workflow syntax validated
- First real release requires dispatching from `main` with `release_version=2.0.0` to confirm end-to-end behavior

[AIML-839]: https://contrast.atlassian.net/browse/AIML-839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/gradle-release.yml
```diff
@@ -3,19 +3,11 @@ name: Gradle Release
 on:
   workflow_dispatch:
     inputs:
-      mode:
-        description: Release flow to run
+      release_version:
```
ChrisEdwards, line 6: User can bump major, minor version here, or leave blank for patch bump .

.github/workflows/gradle-release.yml
```diff
@@ -25,74 +17,7 @@ permissions:
   contents: read
 
 jobs:
-  artifactory-smoke:
```
ChrisEdwards, line 28: Removed all the smoke tests since we know access to artifactory works

.github/workflows/gradle-release.yml
```diff
@@ -131,48 +66,102 @@ jobs:
...
+          if [[ -n "$RELEASE_VERSION" ]]; then
+            tag_name="v${RELEASE_VERSION}"
+
+            if git rev-parse -q --verify "refs/tags/${tag_name}" >/dev/null; then
+              tag_commit="$(git rev-list -n 1 "${tag_name}")"
+              if [[ "$tag_commit" != "$current_head" ]]; then
+                echo "Tag ${tag_name} already exists at ${tag_commit}, not current HEAD ${current_head}"
+                exit 1
+              fi
+              echo "Reusing existing release tag ${tag_name} at current HEAD"
```
ChrisEdwards, line 98: This code path resumes a failed release and reuses the tag if it had been created, but release failed.

.github/workflows/gradle-release.yml
```diff
@@ -288,3 +265,15 @@ jobs:
             echo "Pushing and signing $tag"
             docker push "$tag"
           done
+
+      - name: Create GitHub Release
```
ChrisEdwards, line 269: Moved Github release to the end, so a failure in any other part doesn't result in a successful release.

build.gradle
```diff
@@ -25,30 +35,6 @@ ext.positiveIntProperty = { String propertyName, int defaultValue ->
     }
 }
 
-def requestedNewVersion = providers.gradleProperty('newVersion')
-def gradlePropertiesFile = layout.projectDirectory.file('gradle.properties')
-
-tasks.register('setVersion') {
```
ChrisEdwards, line 31: This is what was causing the problems with protected branch. Gone now

dougj-contrast (approved):

PR #145, opened 2026-06-05
AIML-853 Fix search attack sort syntax

ChrisEdwards:
**Jira:** [AIML-853](https://contrast.atlassian.net/browse/AIML-853)

## Summary
Align the `search_attacks` sort parameter with the `property,DIRECTION` convention used by all other MCP tools, replacing the inconsistent `-prefix` descending syntax.

- AI agents no longer need to learn a one-off sort convention for attacks
- Invalid sort input now produces a clear error message with the expected format
- Codifies the sort convention in MCP_STANDARDS.md for future tools

## Why This Change Exists
The `search_attacks` tool accepted `-startTime` for descending sort while every other tool in the MCP server uses `property,DIRECTION` (e.g., `startTime,DESC`). This inconsistency confused AI agents that had learned the standard convention from other tools, causing unnecessary retries and validation failures. This was finding #7 from the hosted MCP server exploratory testing round.

### Sort parameter parsing (`AttackFilterParams`)
- `AttackFilterParams` now expects `property,DIRECTION` format (e.g., `startTime,DESC`, `status,ASC`)
- Internally converts to the SDK's `-prefix` format so the API contract is unchanged
- Direction is case-insensitive (`desc`, `DESC`, `Desc` all work)
- Whitespace around commas and at edges is trimmed
- Error messages now show the expected format, valid properties, and valid directions

### Tool description (`SearchAttacksTool`)
- `@ToolParam` description updated to document the `property,DIRECTION` format

### MCP Standards (`MCP_STANDARDS.md`)
- Added "Sort Convention" section codifying `property,DIRECTION` as the standard

### Tests
- All unit tests, parity tests, and integration tests updated to use the new format
- New test cases for missing direction, legacy `-prefix` rejection, invalid direction, and too many parts
- Integration tests expanded to cover `status` (bare field without direction) and `-status` (legacy prefix) as invalid inputs

## Testing
- `make check-test` passes (unit + static analysis)
- Integration tests updated and passing with the new sort format
- Parameterized tests cover all five valid sort fields in both ASC and DESC directions

[AIML-853]: https://contrast.atlassian.net/browse/AIML-853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/params/AttackFilterParams.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/params/AttackFilterParams.java
@@ -18,8 +18,11 @@ package com.contrast.labs.ai.mcp.contrast.tool.attack.params;
 import com.contrast.labs.ai.mcp.contrast.sdkextension.data.adr.AttacksFilterBody;
 import com.contrast.labs.ai.mcp.contrast.tool.base.BaseToolParams;
 import com.contrast.labs.ai.mcp.contrast.tool.validation.ToolValidationContext;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 import java.util.Set;
+import java.util.stream.Collectors;
 import lombok.Getter;
 import org.springframework.util.StringUtils;
 
@@ -30,7 +33,7 @@ import org.springframework.util.StringUtils;
  * <p>Usage:
  *
  * <pre>{@code
- * var params = AttackFilterParams.of("ACTIVE", "EXPLOITED", "sql", false, null, null, "-startTime");
+ * var params = AttackFilterParams.of("ACTIVE", "EXPLOITED", "sql", false, null, null, "startTime,DESC");
  * if (!params.isValid()) {
  *   // Handle errors
  * }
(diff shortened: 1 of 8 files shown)


contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/params/AttackFilterParams.java
```diff
@@ -152,6 +141,37 @@ public static AttackFilterParams of(
     return params;
   }
 
+  private static String parseSort(ToolValidationContext ctx, String sort) {
```
Alex-Contrast: nit: sort and return type are `@Nullable` while ctx `@NonNull`

Alex-Contrast (approved):

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/params/AttackFilterParams.java
ChrisEdwards: Addressed in a50a100. `sort` and `parseSort` return/input are now annotated `@Nullable`, `ctx` is annotated `@NonNull`, and the focused nullability contract test plus `./gradlew spotlessCheck checkstyleMain checkstyleTest test` pass.

PR #156, opened 2026-07-13
AIML-1056 Fix CVE CVSS mapping

ChrisEdwards:
**Jira:** [AIML-1056](https://contrast.atlassian.net/browse/AIML-1056)

## Summary

Fix `list_applications_by_cve` so it returns a truthful CVSS score and severity. The tool previously reported `score: 0.0` and blank severity for **every** CVE (Log4Shell included), because the `Cve` model no longer matched TeamServer's CVE-detail response. This restores the severity/score signal agents rely on to prioritize CVE remediation, across both the hosted and stdio servers.

- Nested `cvssv2`/`cvssv3` data now deserializes and is exposed on the `cve` object.
- A single preferred `score`/`severity` is derived (CVSS v3, falling back to v2 severity).
- v2-only CVEs report an absent `score` instead of a misleading `0.0`.
- The legacy flat CVE contract consumed by `get_protect_rules` is preserved.

## Why This Change Exists

TeamServer's `GET /ng/organizations/{orgId}/cves/{cveId}` response nests CVSS metrics under `cvssv2`/`cvssv3` sub-objects and identifies the CVE by a numeric `id`, with no top-level `score` or `references`. Our `Cve` model still declared the old flat, NVD-style shape: a primitive `double score` and top-level CVSS vector strings.

Gson matches field names exactly with no naming policy and no `@SerializedName` remap, so only `name`, `description`, and `status` lined up. The nested `cvssv2`/`cvssv3` objects were silently dropped, and because `score` was a primitive `double`, the missing property defaulted to `0.0`. The net effect: every `list_applications_by_cve` call returned blank severity and a `0.0` score in both the hosted server (`BearerTokenApiClient`) and the stdio app (`SDKExtension`), which share `contrast-mcp-core`. A `0.0` score is worse than missing data — it reads as "no severity." (`status: "unseen"` is unrelated passthrough and does not gate scoring.)

## Approach

- **Reshape `Cve` to the real `CveDetail` shape.** Wire in the already-present-but-unused `CvssV3` model and add `CvssV2`, plus the CVE-intelligence metadata TeamServer actually sends (`id`, `cwe`, `epssScore`, `epssPercentile`, `cisa`, `cvssScoreSource`, `nvdPublished`, `nvdModified`, `firstSeen`).
- **Normalize at the tool seam, not in the model.** `applyPreferredCvssSummary` sets `score`/`severity` from `cvssv3` when present, else `severity` from `cvssv2`. Keeps the model a plain DTO and the preference policy in one place.
- **Represent "no numeric score" as absent.** TeamServer gives no numeric v2 base score, so for v2-only CVEs `score` stays `null` rather than manufacturing `0.0`. `score` was changed from `double` to `Double` to allow this.
- **Nullable wrappers for all new fields** so legacy Protect CVEs (which lack this metadata) don't acquire misleading zeros/`false`.
- **`@JsonInclude(NON_NULL)`** on `Cve` so fields that don't apply to the current response (`score` on v2-only CVEs, and the legacy `uuid`/`references`) are omitted from the agent-facing JSON rather than emitted as `null`.
- The legacy flat CVE fields are retained on the model because `get_protect_rules` still deserializes that shape via Gson.

### CVE model (`contrast-mcp-core/.../sdkextension/data/`)
- `Cve.java` — added nested `cvssv2`/`cvssv3`, numeric `id`, and CVE-intelligence metadata; `score` is now `Double`; `@JsonInclude(NON_NULL)` added. Legacy flat fields retained for the Protect contract.
- `CvssV2.java` (new) — CVSS v2 metric block (vector fields + `severity`), mirroring the existing `CvssV3`.

### Tool response (`.../tool/library/ListApplicationsByCveTool.java`)
- Added `applyPreferredCvssSummary` to derive the preferred `score`/`severity`, called before dedupe.
- Updated the tool description to document the nested v2/v3 metrics and the v2-only absent-`score` caveat.

## Testing

- `CveTest` (new) — asserts Jackson omits unavailable fields (`score`/`uuid`/`references`) for a v2-only CVE, and that Gson still deserializes the **legacy flat** Protect CVE JSON (`uuid`, `score`, `accessVector`, `references`) so `get_protect_rules` is not regressed.
- `ListApplicationsByCveToolTest` (new cases) — parses TeamServer-shaped payloads through the public tool: a v3 CVE yields `score 9.3` / severity `Critical` with all intelligence metadata mapped, and a v2-only CVE falls back to severity `High` with `score` null and v2 vector fields populated.
- `ListApplicationsByCveToolIT` — the live assertion now requires at least one nested CVSS version, a non-blank derived `severity`, and `score` equal to `cvssv3.baseScore` (null when v3 is absent), replacing the old range-only check.
- `make check-test` passed (Spotless + Checkstyle + unit tests).

## Risks / Follow-ups

- `uuid` and `references` are not present in the current CVE-detail response and are now omitted from that path; they remain on the model and continue to populate for the legacy Protect flat shape.
- `make verify` surfaced six pre-existing seeded-data failures in `GetVulnerabilityToolIT` and `SearchAttacksToolIT` that reproduce on clean `origin/main` — unrelated to this change.

[AIML-1056]: https://contrast.atlassian.net/browse/AIML-1056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
@@ -30,10 +30,19 @@ public class Cve {
   private String cwe;
   private Double epssScore;
   private Double epssPercentile;
+
+  /** Whether the CVE appears in CISA's Known Exploited Vulnerabilities catalog. */
   private Boolean cisa;
+
   private String cvssScoreSource;
+
+  /** NVD publication time as Unix epoch milliseconds. */
   private Long nvdPublished;
+
+  /** NVD modification time as Unix epoch milliseconds. */
   private Long nvdModified;
+
+  /** First-seen time as Unix epoch milliseconds. */
   private Long firstSeen;
 
   // Legacy flat CVSS fields retained for the get_protect_rules contract.
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
@@ -24,9 +24,7 @@ import lombok.Data;
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class Cve {
   private Long id;
-  private String availabilityImpact;
   private String name;
-  private String uuid;
   private String description;
   private String status;
   private String cwe;
@@ -37,14 +35,21 @@ public class Cve {
   private Long nvdPublished;
   private Long nvdModified;
   private Long firstSeen;
-  private String severity;
+
+  // Legacy flat CVSS fields retained for the get_protect_rules contract.
+  private String uuid;
   private String accessVector;
   private String accessComplexity;
   private String authentication;
   private String confidentialityImpact;
   private String integrityImpact;
-  private Double score;
+  private String availabilityImpact;
   private List<String> references;
+
   private CvssV2 cvssv2;
   private CvssV3 cvssv3;
+
+  // Preferred summary derived from the nested CVSS data.
+  private Double score;
+  private String severity;
 }
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CvssV2.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CvssV2.java
@@ -15,10 +15,12 @@
  */
 package com.contrast.labs.ai.mcp.contrast.sdkextension.data;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import lombok.Data;
 
 /** Class representing CVSS v2 scoring details for a CVE. */
 @Data
+@JsonInclude(JsonInclude.Include.NON_NULL)
 public class CvssV2 {
 
   private String accessVector;
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CvssV2.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CvssV2.java
@@ -15,36 +15,25 @@
  */
 package com.contrast.labs.ai.mcp.contrast.sdkextension.data;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
-import java.util.List;
 import lombok.Data;
 
-/** Class representing CVE vulnerability information. */
+/** Class representing CVSS v2 scoring details for a CVE. */
 @Data
-@JsonInclude(JsonInclude.Include.NON_NULL)
-public class Cve {
-  private Long id;
-  private String availabilityImpact;
-  private String name;
-  private String uuid;
-  private String description;
-  private String status;
-  private String cwe;
-  private Double epssScore;
-  private Double epssPercentile;
-  private Boolean cisa;
-  private String cvssScoreSource;
-  private Long nvdPublished;
-  private Long nvdModified;
-  private Long firstSeen;
-  private String severity;
+public class CvssV2 {
+
   private String accessVector;
   private String accessComplexity;
   private String authentication;
   private String confidentialityImpact;
   private String integrityImpact;
-  private Double score;
-  private List<String> references;
-  private CvssV2 cvssv2;
-  private CvssV3 cvssv3;
+  private String availabilityImpact;
+  private String exploitability;
+  private String remediationLevel;
+  private String reportConfidence;
+  private String collateralDamagePotential;
+  private String targetDistribution;
+  private String confidentialityRequirement;
+  private String integrityRequirement;
+  private String availabilityRequirement;
+  private String severity;
 }
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
@@ -143,6 +143,7 @@ public class ListApplicationsByCveTool extends SingleTool<ListApplicationsByCveP
       cve.setScore(cve.getCvssv3().getBaseScore());
       cve.setSeverity(cve.getCvssv3().getSeverity());
     } else if (cve.getCvssv2() != null) {
+      cve.setScore(null);
       cve.setSeverity(cve.getCvssv2().getSeverity());
     }
   }
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
@@ -136,10 +136,13 @@ public class ListApplicationsByCveTool extends SingleTool<ListApplicationsByCveP
   }
 
   private static void applyPreferredCvssSummary(Cve cve) {
-    if (cve != null && cve.getCvssv3() != null) {
+    if (cve == null) {
+      return;
+    }
+    if (cve.getCvssv3() != null) {
       cve.setScore(cve.getCvssv3().getBaseScore());
       cve.setSeverity(cve.getCvssv3().getSeverity());
-    } else if (cve != null && cve.getCvssv2() != null) {
+    } else if (cve.getCvssv2() != null) {
       cve.setSeverity(cve.getCvssv2().getSeverity());
     }
   }
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
@@ -17,6 +17,7 @@ package com.contrast.labs.ai.mcp.contrast.tool.library;
 
 import com.contrast.labs.ai.mcp.contrast.client.ContrastApiClient;
 import com.contrast.labs.ai.mcp.contrast.sdkextension.data.App;
+import com.contrast.labs.ai.mcp.contrast.sdkextension.data.Cve;
 import com.contrast.labs.ai.mcp.contrast.sdkextension.data.CveData;
 import com.contrast.labs.ai.mcp.contrast.sdkextension.data.Library;
 import com.contrast.labs.ai.mcp.contrast.sdkextension.data.LibraryExtended;
@@ -60,7 +61,7 @@ public class ListApplicationsByCveTool extends SingleTool<ListApplicationsByCveP
           Takes a CVE ID (e.g., CVE-2021-44228) and returns:
           - apps: List of applications containing vulnerable libraries
           - libraries: The vulnerable library versions
-          - cve: CVE details including severity and description
+          - cve: CVE details including preferred CVSS score/severity and nested v2/v3 metrics
 
           For each application, class usage data is populated:
           - classCount: Total classes in the vulnerable library
@@ -69,6 +70,9 @@ public class ListApplicationsByCveTool extends SingleTool<ListApplicationsByCveP
           Important: If classUsage is 0, the vulnerable library code is likely NOT
           being executed, significantly reducing exploitability risk. Prioritize
           remediation for applications where classUsage > 0.
+
+          score is absent for v2-only CVEs because TeamServer does not provide a numeric v2
+          base score; use severity and cvssv2 metrics in that case.
 
           Related tools:
           - list_application_libraries: Get all libraries for a specific application
@@ -94,6 +98,7 @@ public class ListApplicationsByCveTool extends SingleTool<ListApplicationsByCveP
     if (cveData == null) {
       return null; // SingleTool converts this to notFound response
     }
+    applyPreferredCvssSummary(cveData.getCve());
     dedupeServersById(cveData);
 
     var vulnerableLibs =
@@ -128,6 +133,15 @@ public class ListApplicationsByCveTool extends SingleTool<ListApplicationsByCveP
         apps.size());
 
     return cveData;
+  }
+
+  private static void applyPreferredCvssSummary(Cve cve) {
+    if (cve != null && cve.getCvssv3() != null) {
+      cve.setScore(cve.getCvssv3().getBaseScore());
+      cve.setSeverity(cve.getCvssv3().getSeverity());
+    } else if (cve != null && cve.getCvssv2() != null) {
+      cve.setSeverity(cve.getCvssv2().getSeverity());
+    }
   }
 
   private void enrichAppsWithClassUsage(
--- a/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CveTest.java
+++ b/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CveTest.java
@@ -42,7 +42,14 @@ class CveTest {
         .contains(
             "\"name\":\"CVE-2015-4000\"", "\"severity\":\"High\"", "\"severity\":\"Critical\"")
         .doesNotContain(
-            "\"score\"", "\"uuid\"", "\"references\"", "\"exploitability\"", "\"attackVector\"");
+            "\"score\"",
+            "\"uuid\"",
+            "\"references\"",
+            "\"id\"",
+            "\"epssScore\"",
+            "\"cisa\"",
+            "\"exploitability\"",
+            "\"attackVector\"");
   }
 
   @Test
--- a/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CveTest.java
+++ b/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CveTest.java
@@ -32,12 +32,17 @@ class CveTest {
     var cvssv2 = new CvssV2();
     cvssv2.setSeverity("High");
     cve.setCvssv2(cvssv2);
+    var cvssv3 = new CvssV3();
+    cvssv3.setSeverity("Critical");
+    cve.setCvssv3(cvssv3);
 
     var json = objectMapper.writeValueAsString(cve);
 
     assertThat(json)
-        .contains("\"name\":\"CVE-2015-4000\"", "\"severity\":\"High\"")
-        .doesNotContain("\"score\"", "\"uuid\"", "\"references\"");
+        .contains(
+            "\"name\":\"CVE-2015-4000\"", "\"severity\":\"High\"", "\"severity\":\"Critical\"")
+        .doesNotContain(
+            "\"score\"", "\"uuid\"", "\"references\"", "\"exploitability\"", "\"attackVector\"");
   }
 
   @Test
--- a/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CveTest.java
+++ b/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CveTest.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2025 Contrast Security
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.contrast.labs.ai.mcp.contrast.sdkextension.data;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.contrastsecurity.sdk.internal.GsonFactory;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+
+class CveTest {
+
+  private final ObjectMapper objectMapper = new ObjectMapper();
+
+  @Test
+  void jackson_should_omit_unavailable_cve_fields() throws Exception {
+    var cve = new Cve();
+    cve.setName("CVE-2015-4000");
+    var cvssv2 = new CvssV2();
+    cvssv2.setSeverity("High");
+    cve.setCvssv2(cvssv2);
+
+    var json = objectMapper.writeValueAsString(cve);
+
+    assertThat(json)
+        .contains("\"name\":\"CVE-2015-4000\"", "\"severity\":\"High\"")
+        .doesNotContain("\"score\"", "\"uuid\"", "\"references\"");
+  }
+
+  @Test
+  void gson_should_preserve_legacy_flat_cve_fields_used_by_protect_rules() {
+    var protectData =
+        GsonFactory.create()
+            .fromJson(
+                """
+                {
+                  "success": true,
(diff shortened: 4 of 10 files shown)


contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CvssV2.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.contrast.labs.ai.mcp.contrast.sdkextension.data;
+
+import lombok.Data;
+
+/** Class representing CVSS v2 scoring details for a CVE. */
+@Data
+public class CvssV2 {
```
ChrisEdwards, line 24: **[MEDIUM]** Nested `CvssV2`/`CvssV3` are missing `@JsonInclude(NON_NULL)`, so their null fields leak to the MCP consumer

`@JsonInclude(JsonInclude.Include.NON_NULL)` was added to `Cve`, but a class-level annotation does not cascade into nested objects. There is no global NON_NULL `ObjectMapper` in the tool-output path (Spring AI 1.1.4 serializes results through the default mapper), so a v2-only CVE emits ~8 explicit `"exploitability":null`, `"remediationLevel":null`, ... entries inside `cvssv2` (TeamServer's `CVSSv2.of()` never populates the temporal/environmental metrics). This is exactly the null-noise the parent annotation was meant to eliminate, and `CveTest` only proves omission at the top level.

Also affects: `CvssV3.java` (same missing annotation).

**Suggested fix:** Add `@JsonInclude(JsonInclude.Include.NON_NULL)` to both `CvssV2` and `CvssV3`, and extend `CveTest` to assert nested null-omission.

---
*Found by: Claude (maintainability, testing, correctness, simplicity, API-contract reviewers) — validated against Spring AI serialization path + TeamServer source*

contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveToolIT.java
```diff
@@ -181,9 +177,13 @@ void listApplicationsByCve_should_populate_cve_metadata() {
...
         .as("cve.description must be populated for a known CVE")
         .isNotBlank();
+    assertThat(cve.getCvssv3() != null || cve.getCvssv2() != null)
+        .as("cve must expose at least one nested CVSS version")
+        .isTrue();
+    assertThat(cve.getSeverity()).as("cve.severity must be derived from CVSS data").isNotBlank();
     assertThat(cve.getScore())
-        .as("cve.score must be within CVSS range [%s, %s]", MIN_CVSS_SCORE, MAX_CVSS_SCORE)
-        .isBetween(MIN_CVSS_SCORE, MAX_CVSS_SCORE);
+        .as("cve.score must match cvssv3.baseScore when CVSS v3 is available")
```
ChrisEdwards: **[MEDIUM]** Integration assertion is seed-shaped and dropped the deterministic range invariant

`assertThat(cve.getScore()).isEqualTo(cve.getCvssv3() == null ? null : cve.getCvssv3().getBaseScore())` derives its expected value from the same object under assertion and branches on seed-data shape — the dual-path smell CLAUDE.md warns against (a test should assert a single expected outcome). It also deleted the clean `isBetween(0.0, 10.0)` invariant (and its `MIN_CVSS_SCORE`/`MAX_CVSS_SCORE` constants), so nothing now guards against an out-of-range/negative score reaching the agent. (Note: it is not a pure tautology — deleting production's `setScore` would still turn it red — but it is materially weaker than what it replaced.)

**Suggested fix:** Keep the exact v3 mapping proof in the unit test (already `isEqualTo(9.3)`), and in the IT assert a seed-independent invariant: when `cvssv3` is present, `score` is non-null and `isBetween(0.0, 10.0)`.

---
*Found by: Claude (testing, adversarial, project-standards reviewers) + Gemini*

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveToolTest.java
```diff
@@ -79,6 +132,62 @@ void listApplicationsByCve_should_enrich_class_usage_through_contrast_api_client
...
+            0.99996,
+            true,
+            "NVD",
+            1639008000000L,
+            1716336000000L,
+            1640995200000L);
+  }
+
+  @Test
+  void listApplicationsByCve_should_fallback_to_cvss_v2_severity_when_cvss_v3_is_absent()
```
ChrisEdwards, line 176: **[MEDIUM]** Missing unit coverage for the non-happy paths of `applyPreferredCvssSummary`

Only the v3-happy and v2-fallback paths are covered. Genuinely untested behaviors: (a) a `Cve` with **both** `cvssv2` and `cvssv3` null (the implicit `else` — `score`/`severity` stay null, no throw); (b) a `cvssv3` present without `baseScore` (→ `setScore(null)`); (c) a `cvssv3` present with null `severity` (→ `setSeverity(null)`). These are real, reachable output behaviors of the new code.

**Suggested fix:** Add unit tests for the neither-version-present case (assert `score`/`severity` null, no exception) and the partial-`cvssv3` data variations.

---
*Found by: Claude (testing, adversarial reviewers)*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
```diff
@@ -130,6 +135,15 @@ protected CveData doExecute(ListApplicationsByCveParams params, WarningCollector
     return cveData;
   }
 
+  private static void applyPreferredCvssSummary(Cve cve) {
```
ChrisEdwards, line 138: **[nitpick]** `applyPreferredCvssSummary` repeats the `cve != null` check in both branches

CLAUDE.md prefers guard clauses over nested/duplicated conditionals. Lead with `if (cve == null) return;`, then the branches test only `getCvssv3()` / `getCvssv2()`.

---
*Found by: Claude (project-standards, simplicity, maintainability reviewers)*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
```diff
@@ -130,6 +135,15 @@ protected CveData doExecute(ListApplicationsByCveParams params, WarningCollector
...
+    if (cve != null && cve.getCvssv3() != null) {
+      cve.setScore(cve.getCvssv3().getBaseScore());
+      cve.setSeverity(cve.getCvssv3().getSeverity());
+    } else if (cve != null && cve.getCvssv2() != null) {
```
ChrisEdwards: **[LOW]** The v2 fallback branch never nulls `score`, so the documented contract isn't enforced in code

The `@Tool` description states "score is absent for v2-only CVEs," but this branch only sets `severity`. In practice `score` is already null on this path (TeamServer's `CveDetail` has no top-level `score` — validated), so this is defensive-only, not a live bug. Still worth making the code state the invariant it documents.

**Suggested fix:** Add `cve.setScore(null);` in the `else if (cvssv2 != null)` branch.

---
*Found by: Gemini + Claude (adversarial reviewer) — severity confirmed defensive-only by validation*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
```diff
@@ -15,22 +15,36 @@
...
+  private Long nvdModified;
+  private Long firstSeen;
+  private String severity;
   private String accessVector;
   private String accessComplexity;
   private String authentication;
   private String confidentialityImpact;
   private String integrityImpact;
-  private double score;
+  private Double score;
```
ChrisEdwards: **[LOW]** `score` `double`→`Double` is a published-module signature change and shifts `get_protect_rules` output

`contrast-mcp-core` is a published module, so `getScore()`/`setScore()` changing from primitive to boxed is a source/binary-incompatible signature change for any external consumer compiled against the old artifact (the in-repo consumers recompile, so no practical break). Combined with the new `@JsonInclude(NON_NULL)`, the shared `Cve` also changes `get_protect_rules` output: a Protect CVE with no score now omits `score` entirely instead of emitting `0.0`. Both are arguably improvements, but they are observable changes.

**Suggested fix:** Call this out in the CHANGELOG / release notes (see the docs-staleness note in the review summary).

---
*Found by: Claude (API-contract, correctness reviewers)*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
```diff
@@ -15,22 +15,36 @@
...
   private List<String> references;
+  private CvssV2 cvssv2;
```
ChrisEdwards, line 58: **[nitpick]** The dual-contract nature of `Cve` is undocumented and invites accidental deletion

`Cve` now backs two shapes: the legacy flat CVSS fields (`accessVector`, `confidentialityImpact`, ..., `score`, `uuid`, `references`) used by the `get_protect_rules` path, and the new nested `cvssv2`/`cvssv3` used by applications-by-CVE. The duplication is justified (proven by `gson_should_preserve_legacy_flat_cve_fields_used_by_protect_rules`), but a future reader seeing both flat `accessVector` and `cvssv2.accessVector` may delete the "duplicate" and silently break Protect deserialization. New fields are also interleaved with the legacy flat block (e.g. `availabilityImpact` is orphaned from its siblings).

**Suggested fix:** Group the legacy flat fields together with a one-line comment marking them as the `get_protect_rules` contract; group the derived summary (`score`/`severity`) next to the nested objects.

---
*Found by: Gemini + Claude (maintainability reviewer)*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
```diff
@@ -15,22 +15,36 @@
...
+  private Long id;
   private String availabilityImpact;
   private String name;
   private String uuid;
   private String description;
   private String status;
+  private String cwe;
+  private Double epssScore;
+  private Double epssPercentile;
+  private Boolean cisa;
```
ChrisEdwards, line 35: **[nitpick]** Opaque names on new fields

`cisa` (boolean) conveys neither that it's a flag nor what it asserts (presumably "listed in the CISA Known Exploited Vulnerabilities catalog"); `nvdPublished`/`nvdModified`/`firstSeen` are epoch-millis `Long`s whose names read like flags/counts and carry no unit.

**Suggested fix:** Consider `cisaKev`/`inCisaKev` and a unit hint on the timestamps, each with a `@SerializedName`/`@JsonProperty` alias to preserve the JSON mapping — or at minimum a Javadoc.

---
*Found by: Claude (maintainability reviewer)*

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CveTest.java
```diff
@@ -0,0 +1,85 @@
...
+    cve.setName("CVE-2015-4000");
+    var cvssv2 = new CvssV2();
+    cvssv2.setSeverity("High");
+    cve.setCvssv2(cvssv2);
+
+    var json = objectMapper.writeValueAsString(cve);
+
+    assertThat(json)
+        .contains("\"name\":\"CVE-2015-4000\"", "\"severity\":\"High\"")
+        .doesNotContain("\"score\"", "\"uuid\"", "\"references\"");
```
ChrisEdwards: **[nitpick]** NON_NULL proof omits the newly added nullable fields

The `doesNotContain` check only covers `score`/`uuid`/`references`. The new nullable fields (`id`, `cwe`, `epssScore`, `epssPercentile`, `cisa`, `cvssScoreSource`, `nvdPublished`, `nvdModified`, `firstSeen`) are all null in this fixture but not asserted absent, so a regression that started emitting them as `null` wouldn't be caught.

**Suggested fix:** Add a few representative new fields (e.g. `"id"`, `"epssScore"`, `"cisa"`) to the `doesNotContain` list.

---
*Found by: Claude (testing reviewer)*

contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveToolIT.java
```diff
@@ -181,9 +177,13 @@ void listApplicationsByCve_should_populate_cve_metadata() {
     assertThat(cve.getDescription())
         .as("cve.description must be populated for a known CVE")
         .isNotBlank();
+    assertThat(cve.getCvssv3() != null || cve.getCvssv2() != null)
+        .as("cve must expose at least one nested CVSS version")
```
ChrisEdwards: **[nitpick]** Opaque boolean assertion + severity checked only for non-blank

`assertThat(cve.getCvssv3() != null || cve.getCvssv2() != null).isTrue()` collapses to a bare boolean whose failure message names neither field. Prefer `assertThat(cve).satisfiesAnyOf(c -> assertThat(c.getCvssv3()).isNotNull(), c -> assertThat(c.getCvssv2()).isNotNull())`. Separately, the core behavior here is that `severity` is *derived* from the preferred CVSS version, but the IT only asserts `isNotBlank()` — acceptable if the exact-derivation proof stays in the deterministic unit tests.

---
*Found by: Claude (testing reviewer)*

ChrisEdwards (commented):
## Multi-model review — automated

Consolidated review of this PR by multiple independent engines (Claude review agents across correctness, maintainability/Clean Code, testing, project-standards, simplicity, API-contract, and adversarial/security lenses; OpenAI Codex; Google Gemini; plus a docs-staleness pass). Findings were then adversarially validated against the real TeamServer source to strip false positives.

**Overall: the fix is sound and safe to merge.** The reshaped `Cve` model matches TeamServer's `CveDetail` exactly (all field names verified 1:1 against `origin/develop`), the legacy flat contract for `get_protect_rules` is preserved and tested, and the v3-preferred / v2-severity-fallback behavior is correct. No CRITICAL or HIGH issues survived validation. What remains is one confirmed serialization gap, a weakened integration assertion, a test-coverage gap, and a set of low-severity polish/consistency items.

### Eliminated as false positives (validated, not posted inline)
- **CvssV3 primitive `double` → misleading `0.0` score** (flagged by 6 engines as the headline issue): refuted. TeamServer's `CVSSv3.of()` only builds a `cvssv3` object from a valid vector and always sets `baseScore` from DB column `cvss_3_score` (`nullable=false`), so a present `cvssv3` always has a real `baseScore`; `0.0` only ever means a genuine "None" CVE. The subscore-`0.0` angle is faithful mirroring of TeamServer's own output.
- **Block-granular severity fallback drops an available v2 severity**: refuted. `CVSSv3.of()` unconditionally derives `severity` from `baseScore`, so a present `cvssv3` always carries `severity`.
- **New `Cve`/`CvssV2` fields are speculative (YAGNI)**: refuted. All 9 new `Cve` fields and all 15 `CvssV2` fields map 1:1 to the real TeamServer DTOs; EPSS + CISA-KEV are valuable prioritization signal.

### Documentation staleness (files outside this PR's diff)
- **[MEDIUM] `test-plans/test-plan-list_applications_by_cve.md:111`** — describes `score` as an always-present CVSS score (0.0–10.0) and lists flat `accessVector`/`confidentialityImpact`/etc. as top-level `cve` fields. Stale: `score` is now nullable/absent for v2-only CVEs, and those metrics now live nested under `cvssv2`/`cvssv3`. Add `severity`, `cvssv2`, `cvssv3` and note the v2-only absent-score behavior. *(Docs staleness reviewer)*
- **[LOW] `test-plans/test-plan-list_applications_by_cve.md:44`** — the `CveData` structure comment (`// CVE details (name, score, description)`) understates the reshaped model. *(Docs staleness reviewer)*
- **[LOW] `manual-tests/list-applications-by-cve-manual-test.md:6`** — overview omits the preferred `score` (absent for v2-only) and nested `cvssv2`/`cvssv3` metrics. *(Docs staleness reviewer)*
- **[LOW] `CHANGELOG.md`** — no entry for this observable output change (v2-only CVEs now omit `score` instead of reporting `0.0`; response gains preferred `severity` + nested v2/v3 metrics; `get_protect_rules` CVE output now omits a null `score`). Add an `## [Unreleased]` entry. *(Docs staleness reviewer + Claude API-contract/correctness)*

### Consensus
- Nested `@JsonInclude` gap: 5 engines · IT assertion weakening: 4 engines · guard-clause nit: 3 engines · missing test coverage: 2 engines.

### Verdict
COMMENT — no blocking issues; the inline items are recommended cleanups. (Submitted as a comment review since the author cannot approve their own PR.)

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveToolTest.java
```diff
@@ -79,6 +132,62 @@ void listApplicationsByCve_should_enrich_class_usage_through_contrast_api_client
             });
   }
 
+  @Test
+  void listApplicationsByCve_should_return_cvss_v3_score_when_teamserver_nests_cvss_data()
```
ChrisEdwards, line 138: **[nitpick]** The v3 unit test skips `status` and `description`

The fixture carries `status:"unseen"` and a `description`, but the `containsExactly` extraction omits both, so their deserialization is unverified.

**Suggested fix:** Add `getStatus()` / `getDescription()` to the extraction or a separate assertion.

---
*Found by: Claude (testing reviewer)*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CvssV2.java
ChrisEdwards: Addressed in 4c814c0. Added NON_NULL serialization to both nested CVSS DTOs and extended CveTest to prove absent v2 and v3 fields are omitted. Verified with the focused CveTest task.

contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveToolIT.java
ChrisEdwards: Addressed in cd77bbb. Restored named CVSS v3 range bounds and now asserts that a present v3 score is non-null and between 0.0 and 10.0, leaving exact mapping to the deterministic unit test. The focused integration-test task completed successfully.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveToolTest.java
ChrisEdwards: Partially addressed in 3d270d5. Added focused tests for a CVE with neither CVSS version and for v3 with a null severity; both verify the resulting summary contract. I did not add the proposed null-base-score case because CvssV3.baseScore is a primitive double, so that path cannot produce setScore(null); an absent field becomes 0.0, and the validated TeamServer contract always supplies baseScore when v3 exists. The focused unit test task passes.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
ChrisEdwards: Addressed in 82ac694. applyPreferredCvssSummary now returns immediately for a null CVE, so the version branches no longer duplicate the null check. The focused unit test task passes.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
ChrisEdwards: Addressed in 87f7113. The v2 fallback now explicitly clears score, and its unit test starts with a preexisting score to prove the documented absent-score invariant is enforced. The focused unit test task passes.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
ChrisEdwards: Addressed in 24674d8. The Unreleased changelog now identifies the nullable Double accessor change as breaking and documents the list_applications_by_cve and get_protect_rules output changes.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
ChrisEdwards: Addressed in cdf352e. Cve now groups and labels the legacy flat CVSS fields retained for get_protect_rules, while the nested CVSS objects sit next to their preferred score/severity summary. The focused serialization/legacy-contract tests pass.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/Cve.java
ChrisEdwards: Addressed in 8ae9372 with the minimum-contract option suggested here. I kept cisa, nvdPublished, nvdModified, and firstSeen unchanged because this DTO intentionally mirrors TeamServer JSON names, avoiding aliasing and model drift. Field Javadocs now define CISA-KEV semantics and Unix epoch-millisecond units. The focused Cve test passes.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/CveTest.java
ChrisEdwards: Addressed in e38d83f. The NON_NULL serialization test now explicitly covers representative new metadata fields id, epssScore, and cisa in addition to the existing and nested-field checks. The focused Cve test passes.

contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveToolIT.java
ChrisEdwards: Addressed in 4cdca43. The integration test now uses satisfiesAnyOf with named cvssv3 and cvssv2 assertions, so failures identify the missing nested object. Exact severity derivation remains covered by deterministic unit tests. The focused integration-test task completed successfully.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveToolTest.java
ChrisEdwards: Addressed in f0b250e. The CVSS v3 unit test now includes the fixture description and status in its exact extraction assertion. The focused unit test task passes.

ChrisEdwards:
@ChrisEdwards Follow-up to the [overall review](https://github.com/Contrast-Security-OSS/mcp-contrast/pull/156#pullrequestreview-4689073086): all four documentation-staleness findings are addressed.

* `734dbbc` updates the CveData structure description, documents preferred `severity`, nullable v3-only `score`, and nested `cvssv2`/`cvssv3` metrics in the test plan, and refreshes the manual-test overview.
* `24674d8` adds the Unreleased changelog entry covering the nullable `Double` API signature and observable `list_applications_by_cve` / `get_protect_rules` output changes.

Overall review bodies do not expose a resolvable GitHub thread, so this PR-level comment is the completion reply.

Alex-Contrast (approved):

PR #157, opened 2026-07-13
AIML-1057 Fix filtered route coverage

ChrisEdwards:
**Jira:** [AIML-1057](https://contrast.atlassian.net/browse/AIML-1057)

## Summary

Fix `get_route_coverage` so session-filtered and latest-session calls report accurate coverage counts and route status instead of false zeros.

- Filtered responses now use TeamServer's authoritative `count` / `exercised_count` / `discovered_count` totals.
- Routes missing a status get one derived from their `exercised` timestamp, so no filtered route reports `null`.
- Documents that filtered-route `environments` stays empty because the source omits it.

## Why This Change Exists

Calling `get_route_coverage` with `useLatestSession=true` (or any session-metadata filter) routes to a different TeamServer endpoint (`POST /route/filter`) than the unfiltered call. That filtered response returns correct top-level aggregate totals, but its per-route records omit `status`, `environments`, and `discovered`, carrying only an `exercised` epoch-millis timestamp.

The MCP downstream model discarded those top-level totals and recomputed `exercisedCount`/`discoveredCount` by string-matching each route's `status`. With every filtered route's status null, both counters stayed 0 and `coveragePercent` computed 0.0, even when routes were demonstrably exercised (non-zero `exercised` timestamps). The result was affirmatively wrong output — 0 exercised routes / 0% coverage — that misled agents and users assessing test or branch coverage for a session.

## Approach

Fixed entirely on the MCP side; the counts need no TeamServer change because the filtered response already carries them.

- Deserialize the authoritative top-level counts and prefer them when present, keeping the status-string recomputation only as a fallback for older responses that lack them.
- Derive missing per-route status from data already in hand: non-zero `exercised` → `EXERCISED`, otherwise `DISCOVERED`.
- `environments` cannot be safely invented, so it stays empty on the filtered path and the tool description documents the limitation.

### Response model
- `RouteCoverageResponse.java` — add nullable `count`, `exercisedCount`, `discoveredCount` fields with correct `@SerializedName` Gson mappings so the filtered totals are no longer discarded.

### Mapping
- `RouteMapper.java` — prefer the authoritative response counts over the recomputed ones; add `statusFor(route)` to derive status from the `exercised` timestamp when the source omits it; hoist the status strings to named constants.

### Tool description
- `GetRouteCoverageTool.java` — document that `environments` is empty for session-filtered routes.

## Testing

- `GetRouteCoverageToolTest` — two new unit tests over a fixture shaped like the real filtered response (null route statuses, non-zero `exercised` timestamps, top-level `exercised_count=6`/`discovered_count=42`): one asserts the authoritative counts and 12.5% coverage flow through, the other asserts status is derived (`EXERCISED`/`DISCOVERED`) and environments stay empty.
- `GetRouteCoverageToolIT` — new `assertFilteredCoverageConsistent` helper asserts, on the live filtered and latest-session paths, that `totalRoutes == exercisedCount + discoveredCount == routes.size()`, every status is a known value, and environments are empty.
- `make check-test` — 792 unit tests passed; Spotless and Checkstyle clean.
- `make verify` — route tests passed; six unrelated seeded-data failures in `GetVulnerabilityToolIT`/`SearchAttacksToolIT` reproduce on clean `origin/main`.

## Follow-ups

- Filtered-route `environments` stays empty because TeamServer's `/route/filter` generator omits it. Populating it would require a TeamServer-side change to that generator; left empty and documented as a known limitation for now.

[AIML-1057]: https://contrast.atlassian.net/browse/AIML-1057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/routecoverage/RouteCoverageResponse.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/routecoverage/RouteCoverageResponse.java
@@ -30,7 +30,6 @@ public class RouteCoverageResponse {
   private List<String> messages;
   private List<Route> routes;
 
-  @SerializedName("count")
   private Integer count;
 
   @SerializedName("exercised_count")
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/routecoverage/RouteCoverageResponse.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/routecoverage/RouteCoverageResponse.java
@@ -15,6 +15,7 @@
  */
 package com.contrast.labs.ai.mcp.contrast.sdkextension.data.routecoverage;
 
+import com.google.gson.annotations.SerializedName;
 import java.util.List;
 import lombok.Data;
 
@@ -28,4 +29,13 @@ public class RouteCoverageResponse {
   private boolean success;
   private List<String> messages;
   private List<Route> routes;
+
+  @SerializedName("count")
+  private Integer count;
+
+  @SerializedName("exercised_count")
+  private Integer exercisedCount;
+
+  @SerializedName("discovered_count")
+  private Integer discoveredCount;
 }
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/GetRouteCoverageTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/GetRouteCoverageTool.java
@@ -50,14 +50,23 @@ public class GetRouteCoverageTool
           """
           Retrieves route coverage data for an application.
 
-          Routes can have two statuses:
+          Routes can have these statuses:
           - DISCOVERED: Found by Contrast Assess but has not received any HTTP requests
           - EXERCISED: Has received at least one HTTP request
+          - EXCLUDED: Excluded from route coverage tracking
 
           Response fields:
-          - routes: List of routes with coverage status and details
           - success: Whether the request completed successfully
-          - environments: Empty for session-filtered routes because the source omits this data
+          - messages: Informational messages returned by the service
+          - totalRoutes: Total number of routes
+          - exercisedCount: Number of routes with EXERCISED status
+          - discoveredCount: Number of routes with DISCOVERED status
+          - coveragePercent: Percentage of routes that are exercised
+          - totalVulnerabilities: Total vulnerabilities across the returned routes
+          - totalCriticalVulnerabilities: Critical vulnerabilities across the returned routes
+          - routes: List of routes with coverage status and details. Each route includes
+            environments; this list is empty for session-filtered routes because the source omits
+            that data.
 
           Filtering options (mutually exclusive):
           - No filter: Returns all routes across all sessions
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/GetRouteCoverageTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/GetRouteCoverageTool.java
@@ -57,6 +57,7 @@ public class GetRouteCoverageTool
           Response fields:
           - routes: List of routes with coverage status and details
           - success: Whether the request completed successfully
+          - environments: Empty for session-filtered routes because the source omits this data
 
           Filtering options (mutually exclusive):
           - No filter: Returns all routes across all sessions
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/RouteMapper.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/RouteMapper.java
@@ -22,6 +22,7 @@ import com.contrast.labs.ai.mcp.contrast.sdkextension.data.routecoverage.RouteCo
 import java.util.Collections;
 import java.util.Optional;
 import org.springframework.stereotype.Component;
+import org.springframework.util.StringUtils;
 
 /**
  * Mapper for transforming Route objects into lightweight RouteLight representations. Eliminates
@@ -29,6 +30,9 @@ import org.springframework.stereotype.Component;
  */
 @Component
 public class RouteMapper {
+
+  private static final String STATUS_DISCOVERED = "DISCOVERED";
+  private static final String STATUS_EXERCISED = "EXERCISED";
 
   /**
    * Transform a Route object into a lightweight representation. Removes redundant app data, full
@@ -41,7 +45,7 @@ public class RouteMapper {
     return new RouteLight(
         route.getSignature(),
         Optional.ofNullable(route.getEnvironments()).orElse(Collections.emptyList()),
-        route.getStatus(),
+        statusFor(route),
         route.getRouteHash(),
         route.getVulnerabilities(),
         route.getCriticalVulnerabilities(),
@@ -71,16 +75,19 @@ public class RouteMapper {
     int totalCriticalVulnerabilities = 0;
 
     for (var route : routes) {
-      if ("EXERCISED".equalsIgnoreCase(route.getStatus())) {
+      var status = statusFor(route);
+      if (STATUS_EXERCISED.equalsIgnoreCase(status)) {
         exercisedCount++;
-      } else if ("DISCOVERED".equalsIgnoreCase(route.getStatus())) {
+      } else if (STATUS_DISCOVERED.equalsIgnoreCase(status)) {
         discoveredCount++;
       }
       totalVulnerabilities += route.getVulnerabilities();
       totalCriticalVulnerabilities += route.getCriticalVulnerabilities();
     }
 
-    int totalRoutes = routes.size();
+    int totalRoutes = Optional.ofNullable(response.getCount()).orElse(routes.size());
+    exercisedCount = Optional.ofNullable(response.getExercisedCount()).orElse(exercisedCount);
+    discoveredCount = Optional.ofNullable(response.getDiscoveredCount()).orElse(discoveredCount);
     double coveragePercent =
         totalRoutes > 0 ? Math.round((exercisedCount * 100.0) / totalRoutes * 100.0) / 100.0 : 0.0;
 
@@ -94,5 +101,12 @@ public class RouteMapper {
         totalVulnerabilities,
         totalCriticalVulnerabilities,
         lightRoutes);
+  }
+
+  private static String statusFor(Route route) {
+    if (StringUtils.hasText(route.getStatus())) {
+      return route.getStatus();
+    }
+    return route.getExercised() > 0 ? STATUS_EXERCISED : STATUS_DISCOVERED;
   }
 }
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/RouteMapper.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/RouteMapper.java
@@ -85,6 +85,7 @@ public class RouteMapper {
       totalCriticalVulnerabilities += route.getCriticalVulnerabilities();
     }
 
+    // Filtered responses omit route status, so prefer their authoritative aggregate counts.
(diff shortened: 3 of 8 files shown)


contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/RouteMapper.java
```diff
@@ -71,16 +75,19 @@ public RouteCoverageResponseLight toResponseLight(RouteCoverageResponse response
     int totalCriticalVulnerabilities = 0;
 
     for (var route : routes) {
-      if ("EXERCISED".equalsIgnoreCase(route.getStatus())) {
+      var status = statusFor(route);
```
ChrisEdwards, line 78: **[nitpick]** Aggregation-loop status derivation isn't covered by a deterministic test

Reverting this `statusFor(route)` back to `route.getStatus()` survives every unit test. The two new tests use a fixture whose authoritative `exercised_count`/`discovered_count` override the loop-computed counts (lines 89-90), and the other count-asserting tests set explicit route status — so the "response counts absent + null route status" regime (the filtered case this PR targets) never actually drives an asserted count at this call site. The per-route derivation is covered via the `toRouteLight` path; only this aggregation call site is uncovered.

**Suggested fix:** add a unit test with a response that has no count fields and routes with null status (one `exercised > 0`, one `exercised == 0`), asserting the derived `exercisedCount()` / `discoveredCount()`.

---
*Found by: Claude (testing review), confirmed by validation*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/RouteMapper.java
```diff
@@ -71,16 +75,19 @@ public RouteCoverageResponseLight toResponseLight(RouteCoverageResponse response
...
-      } else if ("DISCOVERED".equalsIgnoreCase(route.getStatus())) {
+      } else if (STATUS_DISCOVERED.equalsIgnoreCase(status)) {
         discoveredCount++;
       }
       totalVulnerabilities += route.getVulnerabilities();
       totalCriticalVulnerabilities += route.getCriticalVulnerabilities();
     }
 
-    int totalRoutes = routes.size();
+    int totalRoutes = Optional.ofNullable(response.getCount()).orElse(routes.size());
```
ChrisEdwards, line 89: **[nitpick]** Count resolution reads as compute-then-discard

The loop above computes `exercisedCount`/`discoveredCount` from per-route status, then these three lines overwrite them (and `totalRoutes`) from the response's authoritative counts. It reads as "compute, then discard," and the three lines look parallel but behave differently — `totalRoutes` is a fresh variable while the other two reassign the loop accumulators. A one-line comment (server aggregate counts are authoritative for the filtered path, which omits per-route status) would make the intent clear.

Note: several engines flagged a potential inconsistency here (mixing authoritative `count` with derived sub-counts → `coveragePercent > 100`). I validated against `teamserver` source that `/route/filter` always returns all three counts together with `discovered = total − exercised`, so that scenario can't occur. This is a readability / defensive-hardening suggestion only, not a bug — optionally resolve the triple all-or-nothing.

---
*Found by: Claude (maintainability review)*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/routecoverage/RouteCoverageResponse.java
```diff
@@ -28,4 +29,13 @@ public class RouteCoverageResponse {
   private boolean success;
   private List<String> messages;
   private List<Route> routes;
+
+  @SerializedName("count")
```
ChrisEdwards: **[nitpick]** `@SerializedName("count")` is redundant

Field `count` already maps to JSON key `count` under the SDK's default (IDENTITY) Gson naming, so this annotation is a no-op. `exercised_count` and `discovered_count` genuinely need theirs (their field names are camelCase). Harmless either way — drop it to match the "annotate only when the key differs" convention used elsewhere in this class, or keep it for symmetry with the two siblings.

---
*Found by: Claude (maintainability review)*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/GetRouteCoverageTool.java
```diff
@@ -57,6 +57,7 @@ public class GetRouteCoverageTool
           Response fields:
           - routes: List of routes with coverage status and details
           - success: Whether the request completed successfully
+          - environments: Empty for session-filtered routes because the source omits this data
```
ChrisEdwards: **[nitpick]** `environments` is documented as a top-level response field, but it's per-route

This bullet sits under "Response fields" alongside `routes`/`success`, implying `environments` is top-level — but it's actually a field on each route object. Consider moving the note into the route-level description.

Separately: this PR makes `totalRoutes`, `exercisedCount`, `discoveredCount`, and `coveragePercent` correct on the filtered path, but the description still doesn't list those aggregate fields (and says routes have "two statuses" while the platform also has EXCLUDED). Worth documenting them here, since this description is the contract surface MCP clients see.

---
*Found by: Gemini + Claude (api-contract review)*

ChrisEdwards (commented):
## Multi-model review — PR #157

Consolidated from four engines (Claude across 6 lenses, Codex, Gemini, docs-staleness) with every CRITICAL/HIGH/MEDIUM finding validated against `teamserver` source and this codebase to strip false positives.

**Verdict: no correctness, security, or contract defects.** The core fix is sound and well-evidenced. Posting as a Comment review (GitHub blocks self-approval). Four low-severity suggestions inline, plus two doc follow-ups below.

### Validated as false positives (dismissed)
The loudest findings did not survive validation against `teamserver`'s `POST /route/filter` implementation (`RouteCoverageAnalyticsRestController` / `ApplicationRouteCoverageAnalyticsJpaDao`):

- **"Mixing authoritative + derived counts yields impossible stats (coveragePercent > 100)"** — *rated HIGH by Gemini.* The endpoint always emits `count`/`exercised_count`/`discovered_count` together, with `discovered = total − exercised` by construction and `exercised` a strict subset of `count`. Impossible stats cannot occur.
- **"IT assertions `totalRoutes == exercisedCount + discoveredCount` and `== routes.size()` are fragile"** — *flagged by Codex + Gemini + Claude (3 engines).* Both hold: the sum identity is algebraic, the MCP client sends no paging, and the count and list queries share one predicate (EXCLUDED routes filtered from both).
- **"`statusFor` mislabels EXCLUDED routes"** — the filtered endpoint filters EXCLUDED routes out entirely, so none reach the mapper.
- **"Vulnerability totals undercount vs authoritative `totalRoutes`"** — no pagination, so `count == routes.size()`.

### Documentation follow-ups (files outside this diff)
- **[MEDIUM]** `manual-tests/get-route-coverage-manual-test.md:343` (Test 21) documents the *pre-fix* behavior — `coveragePercent: 0.0` for a session-filtered query. After this fix it reports ~`50.0` with populated counts. Update the expected result, or a tester will read the corrected output as a regression.
- **[LOW]** `CHANGELOG.md` has no entry for this user-visible fix. Consider an Unreleased "Bug Fixes" note.

### Cross-repo
`contrast-mcp-core` is consumed only by `aiml-services` (hosted MCP server, pinned to 2.0.1). The DTO change is additive and deserialize-only; `RouteCoverageResponseLight` is structurally unchanged. The upgrade is opt-in and safe.

*Engines: Claude (correctness/adversarial, security, maintainability, testing, project-standards, api-contract) · Codex · Gemini · docs-staleness reviewer.*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/RouteMapper.java
ChrisEdwards: Addressed in `2170832`. Added a deterministic `RouteMapperTest` case with response-level count fields unset and two null-status routes (`exercised > 0` and `exercised == 0`). It asserts `totalRoutes`, both derived counts, and `coveragePercent`, so reverting the aggregation-loop `statusFor(route)` call now fails. Verified with `./gradlew :contrast-mcp-core:spotlessApply :contrast-mcp-core:test --tests com.contrast.labs.ai.mcp.contrast.tool.coverage.RouteMapperTest`.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/RouteMapper.java
ChrisEdwards: Addressed in `9e081be`. Added a comment directly above count resolution explaining that filtered responses omit per-route status and their aggregate counts therefore take precedence over the loop-derived fallbacks. I kept the existing per-field fallback behavior rather than adding an all-or-nothing branch because the validated endpoint contract always emits the three aggregate fields together. The focused `RouteMapperTest` suite passes.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/sdkextension/data/routecoverage/RouteCoverageResponse.java
ChrisEdwards: Addressed in `cc4a813`. Removed the no-op `@SerializedName("count")` while retaining the annotations required for `exercised_count` and `discovered_count`. The filtered-response Gson test still deserializes `count` correctly and the full focused `GetRouteCoverageToolTest` suite passes.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/GetRouteCoverageTool.java
ChrisEdwards: Addressed in `2ca0764`. The tool description now documents all aggregate response fields, identifies `environments` as a per-route field (including the filtered-response caveat), and lists `EXCLUDED` alongside `DISCOVERED` and `EXERCISED`. The focused `GetRouteCoverageToolTest` suite passes after the description change.

ChrisEdwards:
@ChrisEdwards Addressed both documentation follow-ups from [this review](https://github.com/Contrast-Security-OSS/mcp-contrast/pull/157#pullrequestreview-4689160625):

* `4d588b1` updates manual Test 21 to expect `totalRoutes: 2`, `exercisedCount: 1`, `discoveredCount: 1`, and `coveragePercent: 50.0`, including the summary row.
* `ab95a85` adds an Unreleased Bug Fixes entry for the corrected session-filtered route coverage aggregates.

Both commits are pushed. This overall review body has no resolvable GitHub thread, so this PR-level comment is the reply.

contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/GetRouteCoverageToolIT.java
```diff
@@ -630,4 +633,20 @@ void getRouteCoverage_unfiltered_should_return_at_least_as_many_routes_as_either
             unfilteredCount, latestSessionCount)
         .isGreaterThanOrEqualTo(latestSessionCount);
   }
+
+  private static void assertFilteredCoverageConsistent(RouteCoverageResponseLight coverage) {
```
Alex-Contrast, line 637: [caution] - idk if TS paginates and it wasn't easy to grep and learn. if responses can be paginated, it might break.

Alex-Contrast (approved):

dougj-contrast (approved):

contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/coverage/GetRouteCoverageToolIT.java
ChrisEdwards: Good instinct to flag this. Confirmed no pagination on this path. The MCP client's `getRouteCoverage` in `ContrastApiClient` takes no `limit`/`offset` (unlike every other call in that interface), so `/route/filter` returns `count` alongside the full route set in a single response, and `totalRoutes == routes.size()` holds by construction (validated against teamserver source in the review body). Added a comment above the assertion in `721f754` documenting that invariant so the next reader doesn't have to grep for it.

PR #159, opened 2026-07-17
AIML-573 Generate and attach release SBOMs

ChrisEdwards:
**Jira:** [AIML-573](https://contrast.atlassian.net/browse/AIML-573)

> [!WARNING]
>  This PR has a failing build due to new requirements for the Wiz Scan. To fix the build I need to update the version of Spring Boot to 4.x. I am doing this now. You can review this PR and I will merge only AFTER that change is merged. Its ticket it AIML-1123.

## Summary
Extends the release workflow to generate Software Bills of Materials for the released jar and Docker image, cryptographically attest the jar SBOM, and attach all SBOMs to the GitHub Release — improving supply-chain transparency for every published version.

- Generates CycloneDX and SPDX SBOMs for both the release jar and the Docker image via Syft
- Adds a GitHub build attestation binding the jar to its CycloneDX SBOM
- Publishes all four SBOM files as GitHub Release assets alongside the jar

## Why This Change Exists
Contrast is committing to supply-chain transparency across its products, and the MCP server had no machine-readable dependency manifest for its releases. Consumers had no way to inspect or verify what shipped inside a given version. Generating SBOMs on release and attaching them to the published artifacts closes that gap and lays the groundwork for the broader public-SBOM initiative.

### `.github/workflows/gradle-release.yml`
- **Install Syft** — pins `anchore/sbom-action/download-syft` by commit SHA (v0.24.0) to install the SBOM scanner.
- **Generate release jar SBOMs** — scans the built jar and emits CycloneDX JSON and SPDX JSON, immediately after the artifact is built.
- **Attest release jar SBOM** — uses `actions/attest-sbom` (v4.1.0) to bind the jar to its CycloneDX SBOM, alongside the existing build-provenance attestation.
- **Generate release Docker image SBOMs** — scans the loaded image (after build, before push) and emits CycloneDX JSON and SPDX JSON.
- **Create GitHub Release** — attaches all four SBOM files to the release in addition to the jar.

## Testing
No automated coverage — this is a release-only CI workflow that cannot run outside a tag push. The action refs are pinned to known-good SHAs, and the SBOM output paths reuse the same `steps.versions.outputs.release` interpolation already proven elsewhere in the job. First real validation happens on the next release run.

- **Publishing to `contrastsecurity.dev/sbom-public/` is not yet implemented.** This PR satisfies the "generate," "attest," and "attach to release" parts of AIML-573; pushing the SBOM to the public SBOM site (AC #2) remains a follow-up pending coordination with the SBOM-site owner on the publishing mechanism.
- SBOM generation runs inline in the release job, so a Syft failure will fail the release. This is intentional — a release should not ship without its SBOM.

[AIML-573]: https://contrast.atlassian.net/browse/AIML-573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -160,13 +160,15 @@ jobs:
           ./gradlew clean :contrast-mcp-stdio-app:bootJar -x test --no-daemon
           test -f "contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar"
 
-      # Pin the Syft version explicitly. The download action SHA-verifies the
-      # downloaded binary against anchore's signed checksums. Bump respecting
-      # the ENTSEC-1742 7-day dependency soak window.
+      # Syft version is intentionally left unset. The download action is
+      # SHA-pinned above, so its baked-in default Syft version is already
+      # deterministic, and the action SHA-verifies the downloaded binary
+      # against anchore's signed checksums. Bumping the pinned action (via
+      # Dependabot) carries the Syft version forward with it; an explicit
+      # syft-version input is not tracked by Dependabot and could silently
+      # go stale.
       - name: Install Syft
         uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
-        with:
-          syft-version: v1.46.0
 
       # Scan the packaged JAR here in both formats for ecosystem interoperability.
       - name: Generate release jar SBOMs
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -176,6 +176,10 @@ jobs:
             test -s "$sbom"
             jq empty "$sbom"
           done
+          jq -e '(.components // []) | length > 0' \
+            "$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json"
+          jq -e '(.packages // []) | length > 1' \
+            "$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json"
 
       - name: Verify tracked worktree is clean before publishing
         run: |
@@ -296,6 +300,10 @@ jobs:
             test -s "$sbom"
             jq empty "$sbom"
           done
+          jq -e '(.components // []) | length > 0' \
+            "$SBOM_DIR/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.cdx.json"
+          jq -e '(.packages // []) | length > 1' \
+            "$SBOM_DIR/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.spdx.json"
 
       - name: Push and sign Docker image with DCT
         env:
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -169,6 +169,12 @@ jobs:
           syft scan file:contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar \
             -o "cyclonedx-json=$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json" \
             -o "spdx-json=$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json"
+          for sbom in \
+            "$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json" \
+            "$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json"; do
+            test -s "$sbom"
+            jq empty "$sbom"
+          done
 
       - name: Verify tracked worktree is clean before publishing
         run: |
@@ -276,6 +282,12 @@ jobs:
           syft scan docker:contrast/mcp-contrast:${{ steps.versions.outputs.release }} \
             -o "cyclonedx-json=$SBOM_DIR/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.cdx.json" \
             -o "spdx-json=$SBOM_DIR/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.spdx.json"
+          for sbom in \
+            "$SBOM_DIR/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.cdx.json" \
+            "$SBOM_DIR/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.spdx.json"; do
+            test -s "$sbom"
+            jq empty "$sbom"
+          done
 
       - name: Push and sign Docker image with DCT
         env:
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -19,6 +19,8 @@ permissions:
 jobs:
   release:
     runs-on: ubuntu-latest
+    env:
+      SBOM_DIR: build/release-sboms
     permissions:
       contents: write
       id-token: write
@@ -163,9 +165,10 @@ jobs:
 
       - name: Generate release jar SBOMs
         run: |
+          mkdir -p "$SBOM_DIR"
           syft scan file:contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar \
-            -o cyclonedx-json=mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json \
-            -o spdx-json=mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json
+            -o "cyclonedx-json=$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json" \
+            -o "spdx-json=$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json"
 
       - name: Verify tracked worktree is clean before publishing
         run: |
@@ -209,7 +212,7 @@ jobs:
         uses: actions/attest-sbom@c604332985a26aa8cf1bdc465b92731239ec6b9e # v4.1.0
         with:
           subject-path: contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar
-          sbom-path: mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json
+          sbom-path: ${{ env.SBOM_DIR }}/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json
 
       - name: Create the trust keys folder
         shell: sh
@@ -271,8 +274,8 @@ jobs:
       - name: Generate release Docker image SBOMs
         run: |
           syft scan docker:contrast/mcp-contrast:${{ steps.versions.outputs.release }} \
-            -o cyclonedx-json=mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.cdx.json \
-            -o spdx-json=mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.spdx.json
+            -o "cyclonedx-json=$SBOM_DIR/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.cdx.json" \
+            -o "spdx-json=$SBOM_DIR/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.spdx.json"
 
       - name: Push and sign Docker image with DCT
         env:
@@ -295,10 +298,10 @@ jobs:
           fail_on_unmatched_files: true
           files: |
             contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar
-            mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json
-            mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json
-            mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.cdx.json
-            mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.spdx.json
+            ${{ env.SBOM_DIR }}/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json
+            ${{ env.SBOM_DIR }}/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json
+            ${{ env.SBOM_DIR }}/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.cdx.json
+            ${{ env.SBOM_DIR }}/mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.spdx.json
           generate_release_notes: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -163,6 +163,7 @@ jobs:
       - name: Install Syft
         uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
 
+      # Scan the packaged JAR here in both formats for ecosystem interoperability.
       - name: Generate release jar SBOMs
         run: |
           mkdir -p "$SBOM_DIR"
@@ -282,6 +283,7 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+      # Scan the loaded image here so both formats describe the exact image that will be pushed.
       # A separate repository produces the registry-bound image SBOM; these copies are release assets only.
       - name: Generate release Docker image SBOMs
         run: |
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -158,6 +158,15 @@ jobs:
           ./gradlew clean :contrast-mcp-stdio-app:bootJar -x test --no-daemon
           test -f "contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar"
 
+      - name: Install Syft
+        uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
+
+      - name: Generate release jar SBOMs
+        run: |
+          syft scan file:contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar \
+            -o cyclonedx-json=mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json \
+            -o spdx-json=mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json
+
       - name: Verify tracked worktree is clean before publishing
         run: |
           git diff --quiet
@@ -195,6 +204,12 @@ jobs:
         uses: actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32 # v4.1.0
         with:
           subject-path: contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar
+
+      - name: Attest release jar SBOM
+        uses: actions/attest-sbom@c604332985a26aa8cf1bdc465b92731239ec6b9e # v4.1.0
+        with:
+          subject-path: contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar
+          sbom-path: mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json
 
       - name: Create the trust keys folder
         shell: sh
@@ -252,6 +267,12 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+      - name: Generate release Docker image SBOMs
+        run: |
+          syft scan docker:contrast/mcp-contrast:${{ steps.versions.outputs.release }} \
+            -o cyclonedx-json=mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.cdx.json \
+            -o spdx-json=mcp-contrast-docker-${{ steps.versions.outputs.release }}.sbom.spdx.json
+
       - name: Push and sign Docker image with DCT
         env:
           DOCKER_CONTENT_TRUST: 1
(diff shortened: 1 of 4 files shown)


.github/workflows/gradle-release.yml
```diff
@@ -252,6 +267,12 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+      - name: Generate release Docker image SBOMs
+        run: |
+          syft scan docker:contrast/mcp-contrast:${{ steps.versions.outputs.release }} \
```
ChrisEdwards, line 304: **[MEDIUM]** Docker image SBOM is generated but never attested

The jar gets both `attest-build-provenance` and `attest-sbom`, but the Docker image SBOM is only generated and attached to the release — there is no `attest-sbom` step for the image, so consumers can't `gh attestation verify` it against the image digest. The `docker push` loop captures no digest, and the build uses `push: false` / `load: true`, so there is currently no registry digest to bind an attestation to.

**Suggested fix:** Add an `attest-sbom` step for the image (capture the pushed digest and use `subject-name`/`subject-digest`), or add a one-line comment documenting that image-SBOM attestation is a deliberate follow-up. Reasonable to defer to a follow-up ticket.

---
*Found by: Claude (security, maintainability, testing) + Gemini (2/3 consensus)*

.github/workflows/gradle-release.yml
```diff
@@ -270,7 +291,12 @@ jobs:
           name: Release ${{ steps.versions.outputs.tag }}
           draft: false
           prerelease: false
-          files: contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar
+          files: |
```
ChrisEdwards, line 337: **[MEDIUM]** Release won't fail if an SBOM file is missing

`step-security/action-gh-release` defaults `fail_on_unmatched_files` to `false`, so if any listed SBOM path is absent (a syft run that exits 0 without producing the file, or a future rename of one of these paths), the release is published without that artifact and only logs a warning.

**Suggested fix:** Add `fail_on_unmatched_files: true` to this step's `with:` block.

---
*Found by: Claude (testing)*

.github/workflows/gradle-release.yml
```diff
@@ -158,6 +158,15 @@ jobs:
           ./gradlew clean :contrast-mcp-stdio-app:bootJar -x test --no-daemon
           test -f "contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar"
 
+      - name: Install Syft
+        uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
+
+      - name: Generate release jar SBOMs
+        run: |
+          syft scan file:contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar \
```
ChrisEdwards, line 177: **[LOW]** SBOMs are written into the repo working-tree root

Two side effects: (1) the later "Verify tracked worktree is clean" check passes only because git ignores untracked files — brittle if that check is ever tightened; (2) these files land in the `context: .` Docker build (`.dockerignore` doesn't exclude `*.sbom.*.json`), changing the `COPY . .` layer hash and busting `cache-from: type=gha` every release.

**Suggested fix:** Write SBOMs to `build/` or `${{ runner.temp }}` (and update the attest/release paths), or add `*.sbom.*.json` to `.dockerignore` and `.gitignore`.

---
*Found by: Claude (correctness, maintainability, testing) + Gemini (2/3 consensus)*

.github/workflows/gradle-release.yml
```diff
@@ -158,6 +158,15 @@ jobs:
...
+            -o cyclonedx-json=mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json \
+            -o spdx-json=mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json
```
ChrisEdwards: **[LOW]** No non-empty / valid-JSON guard after the syft steps

The jar build has a `test -f` guard, but the SBOM steps don't verify their output. `syft scan` exits 0 even when it catalogs zero packages, so a valid-but-empty SBOM could be attested and shipped silently. Low real risk (a Spring Boot fat jar and loaded image clearly have dependencies, and a missing input makes syft exit non-zero), but a cheap defense-in-depth guard.

**Suggested fix:** After each syft step, run `test -s "$f" && jq empty "$f"` for each output file.

---
*Found by: Claude (testing)*

.github/workflows/gradle-release.yml
```diff
@@ -196,6 +205,12 @@ jobs:
         with:
           subject-path: contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar
 
+      - name: Attest release jar SBOM
+        uses: actions/attest-sbom@c604332985a26aa8cf1bdc465b92731239ec6b9e # v4.1.0
+        with:
+          subject-path: contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar
+          sbom-path: mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json
```
ChrisEdwards: **[LOW]** Jar SPDX SBOM is published but not attested

Only the CycloneDX SBOM is bound via `attest-sbom`; the SPDX file (`.sbom.spdx.json`) is attached to the release with no verifiable attestation. Acceptable since both describe the same jar, but consumers might assume parity.

**Suggested fix:** Attest the SPDX file too, or stop publishing it.

---
*Found by: Claude (security)*

.github/workflows/gradle-release.yml
```diff
@@ -158,6 +158,15 @@ jobs:
           ./gradlew clean :contrast-mcp-stdio-app:bootJar -x test --no-daemon
           test -f "contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar"
 
+      - name: Install Syft
+        uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
+
+      - name: Generate release jar SBOMs
```
ChrisEdwards, line 174: **[nitpick]** New SBOM steps have no explanatory comments

The steps encode non-obvious decisions (two formats emitted but only one attested; why generation happens at these specific points). A one-line comment on the jar and docker SBOM steps would help future maintainers.

---
*Found by: Claude (maintainability)*

ChrisEdwards (commented):
## Multi-model review

Reviewed by Claude (correctness, security, maintainability, testing), Codex, and Gemini, plus a documentation-staleness pass. **No blockers — safe to merge.** Actions are SHA-pinned (SHAs verified against their tags), attestation permissions are already present, step ordering is correct, and Codex found nothing. The main theme, flagged independently by Gemini and three Claude reviewers, is that the Docker image SBOM is generated but never attested (the jar is). The rest is cheap hardening and documentation drift.

### Documentation drift (not in this diff, flagged here)
- `RELEASING.md:39` — step 11 reads "Creates the GitHub release and attaches the JAR"; it now attaches the JAR plus four SBOM files. **(MEDIUM)**
- `RELEASING.md:37` — the workflow step-list omits SBOM generation and the jar-SBOM attestation. (LOW)
- `RELEASING.md:87` — the "Verify the Release" checklist doesn't cover the SBOM artifacts. (LOW)
- `CHANGELOG.md` `[Unreleased]` — no entry for SBOMs (precedent: the 2.0.0 "Build provenance attestations" entry). (LOW)

### Noted but out of scope for this PR
- No actionlint/yamllint in PR CI — pre-existing repo-wide gap; this release workflow is `workflow_dispatch`-only, so a YAML/expression error first surfaces on a real release.
- The SBOM/attest steps run after the Artifactory publish, so a failure there yields a partial release — the same non-atomic window as the existing provenance step.

*Found by: Docs staleness reviewer (documentation section)*

.github/workflows/gradle-release.yml
ChrisEdwards: Addressed in 3f21eef. Per the release design, a separate repository owns generation of the registry-bound image SBOM. This workflow now documents that its image SBOM copies are GitHub release assets only, so it deliberately does not add a duplicate image-digest attestation here.

.github/workflows/gradle-release.yml
ChrisEdwards: Fixed in dccedcf by enabling fail_on_unmatched_files on the release action. A missing or renamed JAR/SBOM path will now fail the release step instead of silently omitting the asset.

.github/workflows/gradle-release.yml
ChrisEdwards: Fixed in 87c33fd. All generated SBOMs now live under build/release-sboms, and the attestation/release paths use that directory. The root build directory is already ignored by Git and excluded by .dockerignore, so these files no longer affect the Docker context or rely on untracked root files.

.github/workflows/gradle-release.yml
ChrisEdwards: Fixed in fe8800e. Each Syft invocation is now followed by test -s and jq empty checks for both CycloneDX and SPDX outputs, so empty files and malformed JSON stop the release before attestation or publication.

.github/workflows/gradle-release.yml
ChrisEdwards: Fixed in 8232958. The workflow now creates a second SBOM attestation for the published SPDX document, alongside the existing CycloneDX attestation, both bound to the release JAR.

.github/workflows/gradle-release.yml
ChrisEdwards: Fixed in 7c67368. The workflow comments now explain that both formats are emitted for ecosystem interoperability, that the JAR scan follows packaging, and that the image scan targets the loaded image immediately before publication. The image comment also records the separate ownership of the registry-bound SBOM.

ChrisEdwards:
@ChrisEdwards Addressed the documentation drift from the multi-model review summary in 9544a44: RELEASING.md now lists both SBOM generation phases, both JAR SBOM attestations, all four release assets, and release verification checks; CHANGELOG.md now records release SBOM support. The documentation also records that a separate repository generates the registry-bound image SBOM. GitHub review bodies do not expose a resolvable thread, so there is no thread-resolution action for this item.

.github/workflows/gradle-release.yml
```diff
@@ -158,6 +160,23 @@ jobs:
...
+        run: |
+          mkdir -p "$SBOM_DIR"
+          syft scan file:contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar \
+            -o "cyclonedx-json=$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json" \
+            -o "spdx-json=$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json"
+          for sbom in \
+            "$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.cdx.json" \
+            "$SBOM_DIR/mcp-contrast-${{ steps.versions.outputs.release }}.sbom.spdx.json"; do
+            test -s "$sbom"
+            jq empty "$sbom"
```
ChrisEdwards, line 184: **[MEDIUM]** SBOM validation passes empty / zero-component SBOMs

`test -s` + `jq empty` only prove the file is non-empty and syntactically valid JSON. A CycloneDX/SPDX document with zero components still has metadata, is non-empty, and passes `jq empty` — so a syft mis-catalog (wrong path, or a future regression on nested `BOOT-INF/lib` jars) would silently attest and publish an empty-ish SBOM with no warning. The normal Spring Boot fat-jar path catalogs correctly today; this is a silent-degradation gap, not a guaranteed failure.

Also affects the Docker SBOM validation loop at `.github/workflows/gradle-release.yml:293-298`.

**Suggested fix:** assert minimum content, e.g. CycloneDX `jq -e '.components | length > 0'` and SPDX `jq -e '.packages | length > 1'`; optionally sanity-check that a known dependency (e.g. `spring-core`) appears.

---
*Found by: Claude (adversarial review)*

.github/workflows/gradle-release.yml
```diff
@@ -196,6 +215,18 @@ jobs:
         with:
           subject-path: contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar
 
+      - name: Attest release jar CycloneDX SBOM
```
ChrisEdwards, line 229: **[MEDIUM]** New fallible steps widen the pre-existing release-atomicity window

The Artifactory publish (line 201) and Docker push+sign (line 300) both happen before the GitHub release (line 311, last). This PR adds two `attest-sbom` Sigstore calls here plus a syft Docker scan (line 288) inside that window, and stricter `fail_on_unmatched_files: true` at the final step. A failure after the Artifactory publish leaves a published version with no GitHub release, and a re-run re-runs the Artifactory publish (no skip guard) — which fails if the release repo is immutable.

**This is not a blocker:** the non-atomic pattern is pre-existing (a build-provenance attestation already ran after the Artifactory publish before this PR) and was explicitly deemed out-of-scope in earlier review. This PR only incrementally worsens it.

**Suggested follow-up (separate ticket):** move all fallible local work (SBOM generation + all attestations) before the first irreversible external publish, and/or make the Artifactory publish overwrite-tolerant. Worth confirming whether the release repo allows re-publish, since recovery depends on it.

---
*Found by: Claude (adversarial review)*

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ the correct total, exercised, and discovered route counts and coverage percentag
 
 **Richer CVE scoring details**: `list_applications_by_cve` responses now include nested `cvssv2` and `cvssv3` metrics plus a preferred `severity`. CVSS v3 scores remain available through `score`; v2-only CVEs omit `score` instead of reporting `0.0`. A null `score` is also omitted from `get_protect_rules` CVE output.
 
+**Release SBOMs added**: GitHub releases now include CycloneDX and SPDX JSON SBOMs for both the release JAR and Docker image. Both JAR SBOM formats are bound to the JAR with GitHub attestations.
```
ChrisEdwards, line 28: **[MEDIUM]** SBOM entry categorized under Improvements rather than Security

This is a supply-chain/security feature (SBOMs + cryptographic attestations), but it sits under `### Improvements`. The direct precedent in the `[2.0.0]` block — "Build provenance attestations" and "Docker images signed with DCT" — both live under `### Security`.

**Suggested fix:** add a `### Security` subsection under `## [Unreleased]` (matching the 2.0.0 structure) and move this entry there.

---
*Found by: Claude (project-standards review)*

ChrisEdwards (commented):
## Multi-model review

Safe to merge. This is a well-constructed CI/CD change that adds Syft SBOM generation (CycloneDX + SPDX) for the release JAR and Docker image, binds the two JAR SBOMs to the artifact with GitHub attestations, and attaches all five files to the release with fail-closed behavior (`fail_on_unmatched_files: true`). All 6 prior inline review comments were verified resolved in the actual code, and two external review models found zero issues. No blocking findings survived validation.

Inline comments below are all suggestions (3 MEDIUM, 3 LOW) — none block the merge. The most worthwhile are the SBOM content check and the CHANGELOG categorization.

**Also checked and deliberately not flagged inline:**
- `docs/local-mcp-server.md` (not in this PR) documents only build-provenance attestation and the JAR; it could mention the new JAR SBOM attestations and downloadable SBOM assets. Incomplete, not wrong.
- Inline `${{ steps.versions.outputs.release }}` in `run:` blocks is **not** injectable — the value is regex-anchored to `^[0-9]+\.[0-9]+\.[0-9]+$` on every code path. Optional `env:` indirection would be defense-in-depth only.
- Docker SBOMs are release assets but intentionally **not** attested (documented: a separate repo owns the registry-bound image SBOM). Worth confirming that authoritative SBOM genuinely exists in that repo.
- `SBOM_DIR=build/release-sboms` survives to the release step today because no `gradlew clean` runs after generation; correct as ordered, latent only.

.github/workflows/gradle-release.yml
```diff
@@ -252,6 +283,20 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+      # Scan the loaded image here so both formats describe the exact image that will be pushed.
+      # A separate repository produces the registry-bound image SBOM; these copies are release assets only.
+      - name: Generate release Docker image SBOMs
+        run: |
```
ChrisEdwards, line 301: **[nitpick]** Docker SBOM step relies on the JAR step's `mkdir`

Only the JAR SBOM step runs `mkdir -p "$SBOM_DIR"` (line 169). This Docker step writes into `$SBOM_DIR` with no `mkdir` of its own, and syft does not create missing parent directories. Fine as ordered today, but it breaks if the JAR SBOM step is ever reordered or removed.

**Suggested fix:** add `mkdir -p "$SBOM_DIR"` to this step too.

---
*Found by: Claude (adversarial review + project-standards review)*

.github/workflows/gradle-release.yml
```diff
@@ -252,6 +283,20 @@ jobs:
...
+          syft scan docker:contrast/mcp-contrast:${{ steps.versions.outputs.release }} \
```
ChrisEdwards, line 304: **[nitpick]** Docker image scanned by tag rather than digest (latent risk)

`syft scan docker:contrast/mcp-contrast:${release}` reads the locally-loaded image (the build step uses `load: true`, single-platform) — correct as written. But if anyone later adds `platforms:` (multi-arch images cannot `load`) or removes `load: true`, syft would fall back to *pulling* the tag from DockerHub — which does not exist yet before the push — and fail, or scan a stale image.

**Suggested fix:** scan by immutable digest (`docker:...@${{ steps.build.outputs.digest }}` from the build step output) and/or add a `docker image inspect ... >/dev/null` guard plus a comment noting the single-platform-local-load assumption.

---
*Found by: Claude (adversarial review + correctness review)*

RELEASING.md
```diff
@@ -84,7 +87,8 @@ Then create a GitHub release for the tag and attach `contrast-mcp-stdio-app/buil
 
 ## Verify the Release
 
-- GitHub release exists with the expected tag and attached JAR.
+- GitHub release exists with the expected tag, attached JAR, and CycloneDX and SPDX JSON SBOMs for both the JAR and Docker image.
```
ChrisEdwards, line 92: **[nitpick]** Secondary RELEASING.md sections not updated for SBOMs

This "Verify the Release" checklist was correctly updated to require the four SBOMs, but two sibling sections still describe the JAR only:
- "Testing the Release Workflow in a Fork" step 5 (`RELEASING.md:122`) still says "Verify the release tag and attached JAR" — fork runs do produce SBOMs (Syft needs no secrets).
- The "Manual Release" path (`RELEASING.md:72-86`) has no Syft steps, so a manual release cannot satisfy this four-SBOM checklist.

**Suggested fix:** update the fork-test step to mention SBOMs, and note that the SBOM verify items apply only to the automated workflow path.

---
*Found by: Claude (docs-staleness review + project-standards review)*

.github/workflows/gradle-release.yml
ChrisEdwards: Fixed in 46c9463. Both JAR and Docker validations now require at least one CycloneDX component and more than one SPDX package, in addition to the existing non-empty and valid-JSON checks. The jq predicates were exercised against passing and zero-content examples locally.

.github/workflows/gradle-release.yml
ChrisEdwards: Not changing publication order in this PR. This is a pre-existing, non-blocking pipeline-design concern, and a safe fix first requires confirming whether the Artifactory release repository permits same-version republish and defining idempotency across Artifactory, attestations, Docker DCT, and GitHub releases. I recorded that work in bead br-edv, linked as discovered-from the follow-up review task.

CHANGELOG.md
ChrisEdwards: Fixed in f79e264. The release SBOM and attestation entry now appears under a new Security subsection in Unreleased, matching the existing provenance and Docker-signing precedent.

.github/workflows/gradle-release.yml
ChrisEdwards: Fixed in 62b7bb0. The Docker SBOM step now creates SBOM_DIR itself, so it no longer depends on the earlier JAR SBOM step for directory initialization.

.github/workflows/gradle-release.yml
ChrisEdwards: Fixed in 58b82e3 with the guard-and-comment option. The step now verifies the version tag exists in the local Docker daemon before invoking Syft and explicitly records its dependency on the single-platform load: true build. If that loading behavior changes, the workflow fails at the guard instead of scanning a stale or remote tag. I did not use repository@digest because this image has not been pushed and a build digest is not reliably installed as a local RepoDigest by load: true.

RELEASING.md
ChrisEdwards: Fixed in 515320c. Manual-release documentation now states that its minimal recovery path omits SBOMs and attestations, and the verification checklist is explicitly scoped to automated releases. The fork instructions now verify JAR SBOM generation while accurately noting that an uncredentialed fork stops before Docker SBOM generation and the final GitHub release; a fully configured fork must verify all four attached SBOMs.

.github/workflows/gradle-release.yml
```diff
@@ -158,6 +160,27 @@ jobs:
           ./gradlew clean :contrast-mcp-stdio-app:bootJar -x test --no-daemon
           test -f "contrast-mcp-stdio-app/build/libs/mcp-contrast-${{ steps.versions.outputs.release }}.jar"
 
+      - name: Install Syft
+        uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
```
Alex-Contrast, line 171: [minor] ai callout:
this pins downloader, not syft version.
```
If you want to pin it, download-syft takes a syft-version input, so line 164 becomes a block:
        - name: Install Syft
          uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
          with:
            syft-version: v1.33.0   # pick a current stable Syft release
```

Alex-Contrast (approved):

.github/workflows/gradle-release.yml
ChrisEdwards: Good catch. I will fix.

ChrisEdwards: Good catch, thanks. I dug into it. The download-syft action is already SHA-pinned, so its baked-in default Syft version (currently v1.42.3) is deterministic rather than floating, and the action SHA-verifies the downloaded binary against anchore's signed checksums. An explicit syft-version input would not be tracked by Dependabot, which follows uses refs rather than with input strings, so it could silently go stale while the action moves on. I left it unset and added a code comment explaining the rationale in dc8443f, so the Syft version rides the pinned action's vetted default and moves forward whenever Dependabot bumps the action.

ChrisEdwards: TLDR: The download action hardcodes a specific syft version, so pinning the sha of the downloader also pins the version of syft.

PR #162, opened 2026-07-18
AIML-1135 Add search_servers MCP tool for server inventory and Protect coverage

ChrisEdwards:
**Jira:** [AIML-1135](https://contrast.atlassian.net/browse/AIML-1135)

## Summary

Adds a read-only `search_servers` MCP tool — the first server-domain tool in the MCP server — so AI agents can answer deployment and coverage questions like "which servers are online and do they have Protect enabled?", "where are agents out of date?", and "what's running in production without Protect?". It calls TeamServer's current `POST /ng/{orgId}/servers/filter` endpoint through the existing SDKExtension seam and returns a paginated envelope of ~20-field `ServerSummary` records covering identity, agent health, environment, and Protect/Assess coverage. This closes a gap that previously forced customers into the Contrast UI for the entire server/agent-deployment dimension.

## Why This Change Exists

A Customer Solutions Architect relayed a request from a large ($800K ARR) customer who wants to drive server-inventory and Protect-coverage questions through the MCP server rather than the Contrast UI. The MCP server had zero server-domain tools, so the server/agent-deployment dimension of Contrast — the deployment and coverage backbone where every agent registers a server and Protect licenses are consumed — was entirely invisible to agents. Servers connect applications (many-to-many), Protect licensing, and agent health, so a server tool also unlocks agent chaining with `search_applications` and `search_attacks`. Full converged design (Claude drafted, Codex independently reviewed): `docs/plans/search-servers-tool.md`.

## The Big Picture

The tool follows the established tool-per-class + SDKExtension pattern already used by vulnerability and attack search. The MCP contract is deliberately decoupled from TeamServer wire spellings (`applicationsIds`, `q`, `defend`, `-property` sort) — all of which stay behind the `ContrastApiClient` seam.

```mermaid
flowchart LR
    A["AI agent"] --> B["SearchServersTool<br/>@Tool search_servers"]
    B --> C["ServerFilterParams<br/>validate + translate"]
    C --> D["ContrastApiClient<br/>.searchServers"]
    D --> E["SDKExtension<br/>.getServersFiltered"]
    E --> F["POST /ng/{orgId}/servers/filter"]
    E --> G["ServerSummary<br/>records"]
```

Key design decisions (all verified against TeamServer source):
- **POST filter endpoint over the stock SDK** — the SDK's GET endpoint is deprecated, cannot filter by environment or agent version, and drops the total `count`.
- **Nullable `protectEnabled` (renamed from wire `defend`), never coerced to false** — null means unknown/unavailable; a non-null value is TeamServer's effective state, which may reflect the first visible application's effective config. Coercing null→false would corrupt the central "is Protect on?" question.
- **Single `quickFilter` enum, not decomposed booleans** — fake combinability would force client-side post-filtering and corrupt `totalItems`/`hasMorePages`.
- **Exactly one `expand` value** — TeamServer's expand handling is if/else-if, so `includeApplications=true` sends `expand=applications` only.
- **Locale-independent expand tokens** — the adapter emits hardcoded `"applications"` / `"num_apps"` string constants instead of the SDK `ServerExpandValue` enum's `toString()`. That `toString()` lowercases internally, so under a Turkish or Azerbaijani JVM default locale it produces dotless-ı (`applıcatıons`), silently breaking applications expansion. A locale regression pins this.
- **count=0 fallback** — TeamServer returns `count=0` for any empty page, so an out-of-range page issues one `limit=1,offset=0` fallback to recover the true total.

### 1. Tool + parameter contract — **Focus**
The public MCP surface and where wire translation happens. Start here to judge the contract.
- `tool/server/SearchServersTool.java` (new) — `@Tool` method; filter algebra and null semantics documented in the description.
- `tool/server/params/ServerFilterParams.java` (new) — validation, sort parse/translate, and `toServerFilterBody()` wire translation.
- `result/ServerSummary.java` (new) — the ~20-field record and `fromServer` mapping (nullable Protect / agentOutOfDate derivation).

### 2. SDKExtension wire adapter — **Focus**
Where the contract meets TeamServer. The subtle logic (expand selection, count fallback, envelope normalization) lives here.
- `sdkextension/SDKExtension.java` — `getServersFiltered(...)`: POST body, locale-safe expand selection, count-fallback request.
- `sdkextension/data/server/*.java` (new) — Gson wire DTOs including the `ServersResponseEnvelope` normalization.
- `client/ContrastApiClient.java`, `client/SdkApiClient.java` — the `searchServers` seam method + impl.

### 3. Registration + validation plumbing — **Skim**
- `McpContrastApplication.java` — registers the tool bean.
- `tool/validation/StringSpec.java`, `StringListSpec.java`, `ToolSortParser.java` — small validation-spec additions (including `defaultToQuietly`, which sets a default without emitting a warning so the `quickFilter` default applies before allow-list checking).

## Review notes for the reviewer

This branch went through 4 rounds of multi-model ("GOAT") review (Claude/istari + Codex + Gemini + docs-staleness + validation agents) with 3 fix rounds. Issues found and fixed, each source-verified against TeamServer:
- **R1:** a zero-match `tags` filter was misclassified as an internal error instead of a valid empty result (TeamServer returns `success=false, servers:[], count:0` on that path) — fixed with a narrow, source-verified envelope normalization.
- **R2:** hardcoded `environments` allow-list → derived from the SDK `ServerEnvironment` enum.
- **R3 (HIGH):** `agentOutOfDate` was coerced to `false` instead of `null` because the code used the sentinel string `"NO VERSION AVAILABLE"`, but TeamServer's actual value is `"NA"` (`ServerService.NO_VERSION_AVAILABLE = "NA"`). Fixed to `"NA"` and corrected the spec. Quick-filter allow-list also switched to the SDK `ServerFilterForm.ServerQuickFilterType` enum.
- **R4:** test-quality closeouts plus the `protectPending` investigation, resolved below.
- **R5 (locale hardening):** the applications-expand token was built from the SDK enum's `toString()`, which lowercases with the JVM default locale and corrupts to dotless-ı under `tr`/`az`. Switched to explicit lowercase constants and added locale + contract regression coverage.

**✅ Resolved. `protectPending` is omitted from the contract** (commit `1a93e2f`). TeamServer's `ServerJpaDao.findServersByFilters` projection selects `Server_.defend` into BOTH the `defend` and `defendPending` slots and never selects `Server_.defendPending`, so the wire `defendPending` deterministically mirrors `protectEnabled` and carries no pending-restart information. A field named `protectPending` that echoes enabled state would mislead agents, so the field is removed from `ServerSummary`. The `@Tool` description now states that Protect pending-restart state is not available from this endpoint, and the unmapped wire field keeps an explanatory comment in `ServerDetail`. Filed **[TS-43252](https://contrast.atlassian.net/browse/TS-43252)** for the TeamServer projection defect. It was previously reported in TS-20210 (2022, closed) but no commit references that ticket and the line history shows the DAO fix never landed. Re-adding `protectPending` is backward compatible once TS-43252 is fixed. (`assessPending` is correctly projected and remains in the contract.)

## Testing

`make check-test` (unit + static analysis) passes: 864 tests, 0 failures. Coverage maps to the claims above: filter-body capture tests (translated sort, expand selection, wire-field mapping), null-Protect / unknown-agent mapping that is never coerced, envelope-failure → error not empty success, sanitized HTTP error paths, count-fallback behavior, wire fixtures pinning the first-visible-app overwrite shape, and a Turkish/Azerbaijani-locale regression proving the expand token survives locale-sensitive case folding. Locale-mutating tests set and restore the JVM default locale under a shared resource lock so they cannot corrupt sibling tests running in parallel. `SearchServersToolIT` adds `allSatisfy`-per-filter, boolean-algebra, sort-sweep, and pagination-disjointness integration tests.

Known gap: `make verify` reports `SearchServersToolIT#searchServers_should_filter_by_tag` failing. This is an **intentional fail-loud seeded-data precondition** (per the CLAUDE.md IT standard), not a code defect: the staging org lacks a seeded server tag, so the test throws `AssertionError("requires a seeded server tag")` up front rather than passing vacuously. It needs seeded staging data to go green.

- **Cross-repo source compatibility (out of scope here).** Adding `ContrastApiClient.searchServers` is a source-compatibility requirement for the hosted MCP server (`aiml-hosted-mcp-server`): it must implement and register the method before upgrading its core dependency. Coordinate separately.
- **`protectPending`** omitted from the contract. Re-adding it is backward compatible once TeamServer fixes [TS-43252](https://contrast.atlassian.net/browse/TS-43252).
- Read-only tool: no migrations, feature flags, or data writes.

[AIML-1135]: https://contrast.atlassian.net/browse/AIML-1135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TS-43252]: https://contrast.atlassian.net/browse/TS-43252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ChrisEdwards:
I apologize for the size of this PR. It is a new feature, so that is part of the reason it is large. However, I let the agent loop on reviews, and it created some refactors that should have been their own pr (like extracting sort function across attacks and servers). Anyhow it was too late after I saw that, other commits had landed on top.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/result/ServerSummary.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright 2025 Contrast Security
```
Alex-Contrast: multiple AI reviews and no findings. this is a non-blocking nit, but new files are showing 2025

Alex-Contrast (approved):

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/result/ServerSummary.java
ChrisEdwards: That is odd. I'll fix.

ChrisEdwards: Oh, this is because ALL the files say 2025. I'll create another ticket to update this.

PR #165, opened 2026-07-20
AIML-1145 Add /test-mcp-server pre-release test skill and remove static test plans

ChrisEdwards:
**Jira:** [AIML-1145](https://contrast.atlassian.net/browse/AIML-1145)

## Summary
Adds a `/test-mcp-server` skill that checks the Contrast stdio MCP server before a release, and removes the static test plans it replaces.

- Builds the stdio jar fresh, wires that exact artifact into a separate headless Claude instance, and has that instance test every tool it discovers against a live org.
- Two modes, regular (in-depth, default) and smoke (fast), with an optional focus note.
- Deletes the outdated `test-plans/`, `test-plan-RUNNER-prompt.md`, and `manual-tests/` docs.

## Why This Change Exists
The old test plans were static markdown, one file per tool, and they drifted out of date as tools changed. They also assumed a Contrast MCP server was already configured in the tester's Claude setup, so a run could pass against a stale build while the artifact about to ship was never exercised. We wanted a pre-release check that always tests the real jar and never needs hand-maintained plans.

## Approach
The skill hands the mechanical work to a helper script. It builds the jar, sources credentials from `.env.integration-test` (never printing them, relying on inherited environment so no secret touches disk), and launches one nested headless Opus instance wired to the fresh jar with strict MCP config and a read-only allowlist. That orchestrator asks the running server what tools it exposes and spawns one Sonnet subagent per tool. Each subagent finds its own test data by calling the other tools, then exercises its assigned tool with exploratory testing. Because the tool list comes from the server, a newly added tool is covered with no change to the skill. The prompts are deliberately short and human rather than prescriptive, so the testers use judgement.

Design notes worth calling out for review

- The tool list is discovered at run time, not hardcoded, so coverage tracks the server automatically.
- The run is read-only. Every Contrast tool is a search, get, or list, and the allowlist blocks everything else.
- It is on demand only, before a release or when asked, never a routine dev step. That is stated in the skill and the docs so no agent fires it mid-feature.
- `.claude/` was fully gitignored, so a new rule tracks `.claude/skills/` while keeping local config ignored. Without it the shared skill could not live in the repo.

### New skill
- `.claude/skills/test-mcp-server/SKILL.md`, the runbook and usage.
- `.claude/skills/test-mcp-server/run.sh`, the build-and-drive helper.

### Docs
- `RELEASING.md`, adds a pre-release testing step.
- `CLAUDE.md`, points to the skill under Agent skills and frames it as on-demand only.
- `.gitignore`, tracks `.claude/skills/` and drops the orphaned test-plan ignore entries.

### Removed
- `test-plans/` (14 per-tool plans and README), `test-plan-RUNNER-prompt.md`, and `manual-tests/` (9 docs).

## Testing
The helper passes `bash -n`, the jar builds cleanly, credential sourcing and the temp MCP config were verified, and the `.gitignore` change was confirmed to track the skill while keeping `settings.local.json` ignored. The skill has not had a full live run yet, since it drives a real org and the author wants to review first. Spotless and Checkstyle are Java-only, so the new shell and markdown files fall outside their scope, and no Java changed.

- The first live run is the real validation. Run `smoke` before relying on it for a release.
- The nested run needs `.env.integration-test` pointed at an org with enough varied data, or some tools report inconclusive for lack of data.

[AIML-1145]: https://contrast.atlassian.net/browse/AIML-1145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ChrisEdwards:
I tested this skill and it works great. FOund some issues that I now need to fix :(

ChrisEdwards:
Can anyone take a look at this?

Alex-Contrast:
> I tested this skill and it works great. FOund some issues that I now need to fix :(

I thought this comment indicated you found issues and were working on THIS pr.

Alex-Contrast:
here's what claude said:
```
Ran it. Agreed the exploratory layer earns its place, and it clearly did its job if it already found real bugs.

Two thoughts, neither blocking. build.yml:46 builds the jar and stops there, and the ITs are @SpringBootTest with @Autowired tool beans, so nothing in
CI ever runs the artifact and speaks MCP to it. Registration and schema serialization on the real jar is a genuine gap this closes. That half is
deterministic though, and could be a plain script doing tools/list against a checked-in snapshot plus one known-good call per tool, cheap enough to
run on every commit.

The half that only an agent can test is whether the descriptions are good enough for a model to use the tools correctly, and I think the tester prompt
underuses it. WORKS/ISSUES/BROKEN points at behaviour the ITs already cover, which is where most of the run-to-run variance comes from. Asking each
tester what the description failed to tell it, what it had to guess, and what it got wrong on the first attempt would surface things nothing else here
can, and would be more repeatable.
```

Alex-Contrast (approved):

JacobMagesHaskinsContrast (approved):

ChrisEdwards:
When it does in depth testing it does report back what went wrong and I have it create jira tickets with all the details.

PR #170, opened 2026-07-21
AIML-1164 Add on-demand harness-engineering review and improve commands

ChrisEdwards:
## Why

This repo's AI harness, its CLAUDE.md, its docs/agents routes, commands, hooks, and checks, is what lets a coding agent do real work here. The harness-engineering corpus by Ryan Lopopolo (github.com/lopopolo/harness-engineering) is a read-only bundle of playbooks for improving exactly that, the environment around a fixed model and agent, so the agent can recover intent, operate the real system, prove outcomes, and leave the next run better equipped.

We already used that corpus by hand to run a repository review (AIML-1157 through AIML-1163) and one improve-harness loop, but there was no repo-native entry point, so the usage was ad hoc. This PR gives us a discoverable, just-in-time way to point those playbooks at this repo whenever we want to review or improve our own AI harness, without preloading the corpus into CLAUDE.md.

## What

Two project slash commands, following the corpus's own just-in-time principle.

- `/harness-review` runs the repository-review playbook, a broad read-only diagnostic of whether an agent can recover the job, make a coherent change, prove the outcome, and operate safely here.
- `/improve-harness` runs the improve-harness playbook, one bounded loop from baseline through the smallest owning intervention to a fresh rerun.

Each command locates the corpus, reads its AGENTS.md, follows its routing to the named playbook, and runs it against this repo with the corpus kept read-only.

## How the corpus is pinned

`docs/agents/harness.md` owns a single pinned commit (`226c8d3`) and the locate logic. A developer who already has the corpus checked out can point `HARNESS_ENGINEERING_DIR` at it (used only when it sits on the pin, and never checked out in place, so the corpus stays read-only). Everyone else clones the pinned commit into a per-user cache. Both paths resolve the identical commit, so runs are reproducible across the team. Both were tested end to end.

## Also in this PR

Standardizing the `.claude/` gitignore to the Claude Code convention. The repo previously ignored all of `.claude/`. It now ignores only the personal `settings.local.json` and the transient `worktrees/`, so the project `settings.json`, the two `hooks/` scripts, and the new `commands/` are shared with the team. The settings.json is a permission allowlist plus hook wiring, no secrets.

## Non-goals

The corpus is not embedded here and not pasted into CLAUDE.md. Its file layouts, version pins, and fixtures are not copied in. Packaging the commands as a marketplace plugin so every Contrast repo gets them is noted as a follow-up in docs/agents/harness.md.

## Testing

No application code changed, so there is nothing new to unit test. The only executable logic is the corpus-locate snippet in `docs/agents/harness.md`, verified on both the local-checkout and clone-on-demand paths, each resolving the pinned commit.

Jira ticket AIML-1164.

### DIFF
--- a/.claude/hooks/checkstyle.sh
+++ b/.claude/hooks/checkstyle.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+# PostToolUse hook: run Checkstyle after Java files are created or edited
+
+f=$(jq -r '.tool_input.file_path // ""')
+
+# Only run for Java files
+if ! echo "$f" | grep -qE '\.java$'; then
+  exit 0
+fi
+
+# Navigate to project root (two levels up from .claude/hooks/)
+cd "$(dirname "$0")/../.." || exit 0
+
+output=$(./gradlew checkstyleMain checkstyleTest 2>&1)
+rc=$?
+
+[ $rc -eq 0 ] && exit 0
+
+# Feed violations back to Claude as additional context
+jq -n --arg ctx "Checkstyle violations after editing $f:
+
+$output" \
+  '{"hookSpecificOutput":{"hookEventName":"PostToolUse","additionalContext":$ctx}}'
--- a/.claude/commands/harness-review.md
+++ b/.claude/commands/harness-review.md
@@ -1,0 +1,42 @@
+---
+description: Run the harness-engineering repository-review playbook against this repo, corpus kept read-only.
+argument-hint: [optional focus, e.g. "delivery and authority"]
+---
+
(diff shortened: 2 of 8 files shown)


.claude/hooks/checkstyle.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
ChrisEdwards, line 1: I thought this was alredy checked in...but my `.claude` folder was ignored. This was created in the training video.

Alex-Contrast (approved):

PR #171, opened 2026-07-22
AIML-1158 Give vulnerability status a typed owner and render tool-param text from it

ChrisEdwards:
## What & why

The vulnerability **status** vocabulary was stringly-typed and split across three
places that had already drifted apart:

- `ValidationConstants.VALID_VULN_STATUSES` accepted **7** values (including
  `NotAProblem` and `AutoRemediated`)
- the `@ToolParam` the agent actually reads advertised only **5**
- the param javadoc listed **7**

An agent building a query from the advertised set silently **under-filtered** —
a live defect in the tool contract, which is the product's real compatibility
surface (AIML-1158).

## Change

Give the vocabulary a single typed owner: a new core-owned **`VulnStatus`** enum.
Each constant carries its exact TeamServer API value, and the type owns:

- `validApiValues()` — the allowed set for validation
- `defaultApiValues()` — the actionable default subset (`Reported, Suspicious, Confirmed`)
- `PARAM_DESCRIPTION` — the model-facing `@ToolParam` text
- `DEFAULT_REASON` — the warning surfaced when the default subset is applied

Both `SearchVulnerabilitiesTool` and `SearchAppVulnerabilitiesTool` now render
their status description from `VulnStatus.PARAM_DESCRIPTION`; both params classes
validate and default from `VulnStatus`. `ValidationConstants.VALID_VULN_STATUSES`
and `DEFAULT_VULN_STATUSES` are removed.

## Contract decision: widened to all 7 statuses

The design flagged one thing to confirm against the product: does the TeamServer
trace-filter API actually accept `NotAProblem` and `AutoRemediated` as filter
inputs? Confirmed yes:

- a live `search_vulnerabilities` call with `statuses=NotAProblem,AutoRemediated`
  returned `success` with results and no errors
- `platform-public-api` `V4VulnerabilityStatus` maps both values; Contrast PS
  scripts filter `/orgtraces/filter` on them

So the enum advertises the full 7-value set rather than narrowing.

## Tests

- **`VulnStatusTest`** — including an **anti-drift guard** asserting the statuses
  advertised in `PARAM_DESCRIPTION` equal `validApiValues()` exactly (both
  directions). This closes the loop that allowed the original 5-vs-7 drift.
- New param-level tests proving `NotAProblem`/`AutoRemediated` validate and reach
  the filter body's status set.
- Updated the affected default-warning assertions across the parity/IT tests to
  the accurate widened reason text.

## Verification

- `make check-test` — green (875 unit tests, static analysis clean)
- `make verify` — all unit tests and **all 4 vulnerability integration test
  classes pass** (Get/ListTypes/SearchApp/Search). Two unrelated IT failures are
  environmental and pre-existing, not regressions from this change:
  - `ListApplicationLibrariesToolIT` — transient **503 first byte timeout** from
    the Contrast API (library tool)
  - `SearchServersToolIT#searchServers_should_filter_by_tag` — missing seeded
    server-tag data in the test org (server tool)

## Out of scope

Severity and environment typing is deliberately deferred (it touches the SDK
request-body boundary and needs its own design decision). See the bead design
field for the follow-up outline.

Closes AIML-1158 · bead mcp-3rt4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ the correct total, exercised, and discovered route counts and coverage percentag
 
 ### Improvements
 
+- search_vulnerabilities / search_app_vulnerabilities now advertise all 7 valid statuses (adds NotAProblem, AutoRemediated, previously accepted but undocumented); default-exclusion warning text updated. (AIML-1158)
+
 **New `search_servers` tool**: Search the EAC-visible server inventory by agent health,
 environment, application, tag, and Protect coverage with pagination and sorting.
 
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/validation/ValidationConstants.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/validation/ValidationConstants.java
@@ -15,9 +15,6 @@
  */
 package com.contrast.labs.ai.mcp.contrast.tool.validation;
 
-import java.util.List;
-import java.util.Set;
-
 /** Shared validation constants for MCP tools. */
 public final class ValidationConstants {
 
@@ -38,21 +35,6 @@ public final class ValidationConstants {
 
   /** Minimum page number (1-indexed). */
   public static final int MIN_PAGE = 1;
-
-  /** Valid vulnerability status values. */
-  public static final Set<String> VALID_VULN_STATUSES =
-      Set.of(
-          "Reported",
-          "Suspicious",
-          "Confirmed",
-          "NotAProblem",
-          "Remediated",
-          "Fixed",
-          "AutoRemediated");
-
-  /** Default vulnerability statuses (actionable only, excludes Fixed/Remediated). */
-  public static final List<String> DEFAULT_VULN_STATUSES =
-      List.of("Reported", "Suspicious", "Confirmed");
 
   private ValidationConstants() {}
 }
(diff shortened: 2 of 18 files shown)


ChrisEdwards:
## Multi-model review completed (1 round) + fixes applied

Ran a GOAT-style multi-engine review (Claude across 6 lenses, Codex, Gemini, docs-staleness). **No CRITICAL/HIGH; security, correctness, and Codex all clean.** The runtime-accepted status set was already 7 on `main` — only the advertised `@ToolParam` text moved 5→7, closing the documented contract gap.

MEDIUM findings were all addressed in `7b14955`:

- **M1 — anti-drift gap:** the exclusion prose (`PARAM_DESCRIPTION` "excludes …" clause, `DEFAULT_REASON`) sat outside the anti-drift guard. Fixed by deriving `DEFAULT_REASON` from the enum-owned `nonDefaultApiValues()` (valid minus default) so it can't drift, and adding guards that assert both the advertised CSV **and** the exclusion clause equal the computed sets exactly.
- **M2 — duplication:** both params classes now go through `VulnStatus.resolveStatuses(ctx, statusesParam)` instead of copying the validate/default chain.
- **M3 — dropped coverage:** restored the immutability regression test for `defaultApiValues()` (plus `nonDefaultApiValues()`); collections hoisted to `static final`.
- **M4 — stale docs:** refreshed both test plans, the manual test guide, and `CHANGELOG.md` for the 7-status contract and current warning text.

Also folded in the cheap related LOWs: collection caching (L1), list-based drift parser (L3), and pinning the param warning tests to `VulnStatus.DEFAULT_REASON` (L4). Deterministic error-message ordering falls out of the enum-ordered set (L2). Left as-is: `VulnStatus` package placement (L5, legal in-module arrangement, matches the bead design path) and MCP-schema-level `@ToolParam` verification (L6, compile-enforced).

**Verification:** `make check-test` green (880 tests, 0 failures, static analysis clean). Earlier `make verify` passed all vulnerability integration tests; the only 2 IT failures were unrelated environmental issues (a transient 503 in the library tool, missing seeded server-tag data in the server tool).

ChrisEdwards:
Closing — the solution over-abstracted the problem. Will reattempt with a simpler approach.

PR #173, opened 2026-07-27
AIML-501 Add JaCoCo code coverage with enforced floors and CI gating

ChrisEdwards:
**Jira:** [AIML-501](https://contrast.atlassian.net/browse/AIML-501)

## Summary
Adds JaCoCo coverage measurement to the Gradle build with per-module floors that are actually enforced, wired into `check`, `make check-test`, and CI.

- The repo measured no coverage at all before this. There was no JaCoCo plugin, no coverage tasks, and nothing in CI.
- Floors are enforced in-repo by `jacocoTestCoverageVerification`, which `check` now depends on, so they cannot be skipped.
- CI gates on coverage, publishes a summary table to the job summary, and uploads the HTML and XML reports as an artifact.

## Why This Change Exists
AIML-501 was filed in February and sat in Backlog. It was written against Maven, naming `pom.xml`, the `prepare-agent` and `report` goal bindings, and `target/site/jacoco`. None of that applies now that the repo is Gradle, so the ticket and its bead were rewritten for Gradle before the work started.

The gap this closes is that nothing measured coverage. A test could be deleted and no build step would notice.

The sibling repo `aiml-services` has coverage, but its in-repo JaCoCo minimum is set to `0.0` and the real 85 percent lives only in an external `@master`-pinned action plus a skippable pre-push hook. AIML-1170 tracks that as a bug. This PR deliberately avoids repeating it.

## Approach
Floors are set per module, just below the measured baseline, so they block regression without failing on adoption.

| module | line | branch | line floor | branch floor |
| --- | --- | --- | --- | --- |
| contrast-mcp-core | 92.2% | 82.5% | 90% | 80% |
| contrast-mcp-stdio-app | 75.3% | 69.2% | 72% | 65% |

Three alternatives were rejected.

A single combined gate at 85 percent was rejected because combined line coverage is already 88.6 percent, so it would pass on day one and enforce nothing, and core's high coverage would mask rot in stdio-app.

The `aiml-services` shared action `Contrast-Security-Inc/common-github-actions/test-coverage@master` cannot be reused. That repo is private under Contrast-Security-Inc and this one is public under Contrast-Security-OSS, so the default `GITHUB_TOKEN` cannot resolve it and fork PRs never could. Enforcement therefore lives in Gradle, which runs everywhere including forks, and the reporting is done with a SHA-pinned `upload-artifact` rather than an unpinned reference.

The `org.barfuin.gradle.jacocolog` plugin that `aiml-services` uses for its console summary was rejected in favour of a 40-line native `coverageSummary` task, to avoid adding a third-party build dependency to a public repo for formatting alone.

### Build configuration
- `gradle.properties` pins `jacocoVersion=0.8.15`, published 2026-06-04 and so past the ENTSEC-1742 seven-day soak window. No inline or dynamic version.
- `build.gradle` applies `jacoco` across `subprojects`, enables XML alongside Gradle's default HTML, makes `test` finalize with `jacocoTestReport`, and configures `jacocoTestCoverageVerification` with the floors above before wiring it into `check`.
- `build.gradle` declares `ext.coverageMinimums` and `ext.coverageExcludedClassFiles` at the root, and fails the build with an actionable message if a module has no declared floor.
- `build.gradle` registers `coverageSummary`, which parses the JaCoCo XML and emits a Markdown table that reads correctly both in a terminal and when appended to the CI job summary.

### CI
- `.github/workflows/build.yml` adds `jacocoTestCoverageVerification` to the existing Gradle line, appends the summary to `GITHUB_STEP_SUMMARY`, and uploads reports via `actions/upload-artifact` pinned to `043fb46` (v7.0.1). Both new steps use `if: always()` so a coverage failure still publishes the evidence needed to diagnose it.

### Developer workflow and docs
- `Makefile` adds a `coverage` target and includes it in `check-test`.
- `CLAUDE.md` and `CONTRIBUTING.md` document the target, the report paths, and the rule that floors get raised as coverage improves and never lowered to make a build pass.

`McpContrastApplication` is the only excluded class, being `main()` plus bean wiring that unit tests do not exercise.

## Testing
`make check-test` passes from clean, with 870 unit tests, 0 failures, static analysis clean, and both modules over their floors.

The gate was mutation-tested rather than assumed. Temporarily raising the stdio-app line floor to `0.99` failed the build with `Rule violated for bundle contrast-mcp-stdio-app: lines covered ratio is 0.75, but expected minimum is 0.99`. This matters because a coverage gate that silently enforces nothing is the exact defect AIML-1170 describes.

`coverageSummary` output was cross-checked against an independent parse of the same XML reports, and the figures agree to the decimal.

The exclusion was verified by class count. stdio-app drops from 12 classes to 11, with line coverage moving 73.9 percent to 75.3 percent, confirming the filter takes effect on both the report and the enforced floor.

`.github/workflows/build.yml` was parsed to confirm valid YAML, and every action in it resolves to a 40-character SHA.

**Honest gap.** There are no automated tests for the build logic itself. The floors and the summary parser are Groovy in `build.gradle`, and testing them properly needs a `buildSrc` module with a GradleRunner harness, which is what `aiml-services` has for its Kotlin equivalent. Both were verified by hand as described above, but that is not a regression test.

**`make verify` does not pass on this branch, and it does not pass on `main` either.** Three integration tests fail. To check whether this PR caused them, the entire change was stashed and the tests were re-run against clean `main`, where two fail identically with no JaCoCo present.

- `ListApplicationLibrariesToolIT.listApplicationLibraries_should_populate_version_and_staleness_fields` asserts `monthsOutdated >= 0`, but TeamServer returns -112 for `doctrine/lexer` and -59 for `doctrine/annotations`, both reporting a `latestVersion` older than the installed version. Either a TeamServer data problem or a missing guard in the mapper.
- `SearchServersToolIT.searchServers_should_filter_by_tag` fails on its own seeded-data precondition. The test org is missing a server tag.
- `GetRouteCoverageToolIT` failed once on an HTTP 503 Varnish first-byte timeout and passed on re-run, so it is transient.

These are unrelated to coverage and need their own tickets. The repo rule is no PR until `make verify` passes, so this is flagged rather than glossed over.

Two follow-ups worth tracking separately. The stdio-app floors of 72 and 65 percent are a starting ratchet and should climb, with `SDKHelper` at 43 percent line coverage over 114 lines being the main drag. Separately, this branch began as a spike evaluating the CRAP metric via `crap-java`, which surfaced that `ServersResponseEnvelope` has zero test coverage despite encoding a subtle TeamServer tag-prefilter workaround where a wrong condition silently converts a genuine API failure into an empty result.

[AIML-501]: https://contrast.atlassian.net/browse/AIML-501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,5 +29,21 @@ jobs:
         java-version: '21'
         distribution: 'temurin'
         cache: gradle
-    - name: Run checks, tests, and build
-      run: ./gradlew spotlessCheck checkstyleMain checkstyleTest test :contrast-mcp-stdio-app:bootJar --no-daemon
+    - name: Run checks, tests, coverage, and build
+      run: ./gradlew spotlessCheck checkstyleMain checkstyleTest test jacocoTestCoverageVerification :contrast-mcp-stdio-app:bootJar --no-daemon
+
+    # Runs even when the coverage floor fails, since that is exactly when the
+    # numbers and the HTML report are needed to diagnose the failure.
+    - name: Publish coverage summary
+      if: always()
+      run: ./gradlew --quiet coverageSummary --no-daemon >> "$GITHUB_STEP_SUMMARY"
+
+    - name: Upload coverage reports
+      if: always()
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: jacoco-coverage-reports
+        path: |
+          **/build/reports/jacoco/test/html/**
+          **/build/reports/jacoco/test/jacocoTestReport.xml
+        if-no-files-found: warn
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,9 @@ jobs:
       run: ./gradlew spotlessCheck checkstyleMain checkstyleTest test jacocoTestCoverageVerification :contrast-mcp-stdio-app:bootJar --no-daemon
 
     # Runs even when the coverage floor fails, since that is exactly when the
-    # numbers and the HTML report are needed to diagnose the failure.
+    # numbers and the HTML report are needed to diagnose the failure. coverageSummary
+    # only reads the reports the step above wrote, so this does not re-run the tests;
+    # if that step failed before producing them, the table says so per module.
     - name: Publish coverage summary
       if: always()
       run: ./gradlew --quiet coverageSummary --no-daemon >> "$GITHUB_STEP_SUMMARY"
@@ -47,3 +49,5 @@ jobs:
           **/build/reports/jacoco/test/html/**
           **/build/reports/jacoco/test/jacocoTestReport.xml
         if-no-files-found: warn
+        # Diagnostic output for a failed run, not a durable artifact.
+        retention-days: 7
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,11 +70,15 @@ This repository uses the Gradle wrapper and Java 21. Do not use Maven commands o
 **Common commands:**
 ```bash
 ./gradlew spotlessCheck checkstyleMain checkstyleTest test
+./gradlew jacocoTestCoverageVerification coverageSummary
 ./gradlew :contrast-mcp-stdio-app:bootJar
 ./gradlew :contrast-mcp-core:publishToMavenLocal :contrast-mcp-core:verifyCorePublicationMetadata
+make coverage
 make check-test
 make verify
 ```
+
+**Coverage floors:** JaCoCo measures coverage on every `test` run. Per-module floors live in `ext.coverageMinimums` in the root `build.gradle` and are enforced by `jacocoTestCoverageVerification`, which `check` and `make check-test` both run, so a coverage regression fails the build. Raise a floor as coverage improves. Never lower one to make a build pass, and treat additions to `coverageExcludedClassFiles` like a checkstyle suppression.
 
 **Module split:**
 - `contrast-mcp-core` is the transport-neutral shared library published as `com.contrast.labs.ai.mcp:contrast-mcp-core`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -36,7 +36,8 @@ Use these make targets for all checks and tests:
 ```bash
 make check       # Auto-format then run static analysis (no need to run make format first)
 make test        # Run unit tests (quiet output)
-make check-test  # Run static analysis and unit tests
+make coverage    # Verify JaCoCo coverage floors and print the summary
+make check-test  # Run static analysis, unit tests, and coverage
 make verify      # Run all tests including integration
 make format      # Auto-format code with Spotless (also runs automatically via make check)
 make build       # Build the project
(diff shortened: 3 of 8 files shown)


Alex-Contrast (approved):

dougj-contrast (approved):
## Java PR Review: AIML-501 — Add JaCoCo code coverage with enforced floors and CI gating

Independent review pass (Claude + Codex CLI + Copilot CLI) against this diff. Well-built, unusually carefully-reasoned build change — nothing blocking.

### Important
- **`make check-test` runs the stdio-app's unit test suite twice.** `check-test` chains `make check`, `make test`, `make coverage` as three separate `./gradlew` invocations. `coverage`'s `jacocoTestCoverageVerification` transitively depends on `test` via `jacocoTestReport`, and per the PR's own comment, `contrast-mcp-stdio-app`'s `test` task is never UP-TO-DATE across invocations (`bootBuildInfo` rewrites `build-info.properties` every run). So the standalone `make test` step and the transitive `test` inside `make coverage` both re-run that module's suite, roughly doubling its portion of every local `check-test` run. **CI is unaffected** — `build.yml` lists `test jacocoTestCoverageVerification` in one single gradle invocation, so task-graph dedup handles it there. Worth combining `test` and `coverage` into one invocation in the Makefile, or accepting the cost with a comment.
- **Residual per-counter NaN gap in the coverage gate.** The new `doFirst` guard correctly catches "no execution data at all" and "no class files at all" — the aggregate version of the exact defect this PR closes (JaCoCo's `Limit.check` reports no violation when a counter's total is zero, since the ratio is `0/0 = NaN`). But it doesn't guard per-counter: if a module ever has zero branch positions, the `BRANCH` floor alone would silently pass while `LINE` still enforces. Both modules have branches today so this is dormant, not exploitable now — worth at least a comment acknowledging the gap given how carefully the rest of this guard is reasoned through.

### Code smells (optional polish)
- `if: always()` on the two new CI steps also runs on workflow cancellation, keeping a cancelled job alive. `if: ${{ !cancelled() }}` preserves "publish even on failure" while still letting cancellation actually cancel.
- The "No coverage minimums declared" `GradleException` is duplicated verbatim in `jacocoTestCoverageVerification` and `coverageSummary` — a shared closure would give it one source of truth.
- `test` is redundant on the CI gradle command line (already pulled in transitively by `jacocoTestCoverageVerification`).
- The `doFirst` block interpolates `${project.name}` at execution time — not a problem today (config cache isn't enabled), but `coverageSummary` already shows the safer pattern (capturing locals at configuration time) if that ever changes.
- Two near-identical `mustRunAfter subprojects.collect{...}` lines in `coverageSummary`, combinable into one.

### Good
- XXE hardening in `coverageSummary`'s `XmlSlurper` setup is textbook-correct defense-in-depth (DOCTYPE allowed since JaCoCo's XML declares one, but external DTD/entities disabled, `secure-processing` re-enabled) — well-reasoned even for trusted build output.
- `setOnlyIf { true }` + explicit `doFirst` preconditions correctly neutralizes Gradle's built-in silent-skip-when-no-exec-data behavior — exactly the AIML-1170 defect class, and mutation-tested per the PR description.
- `coverageSummary`'s deliberate `mustRunAfter` (not `dependsOn`) correctly avoids the `bootBuildInfo` UP-TO-DATE trap, and is documented precisely.
- Shared `coverageClassDirs` keeps the report and the enforced verification from ever drifting out of sync.
- Checked independently: the `actions/upload-artifact` SHA matches the `v7.0.1` tag exactly, and `jacocoVersion=0.8.15` was genuinely published in early June — well past the 7-day ENTSEC-1742 soak window.

Approving — nothing here blocks merge; the two "Important" items are worth a look but neither affects correctness of the CI gate itself.

ChrisEdwards:
Thanks @dougj-contrast, this was a genuinely useful pass. All seven findings are addressed.

They landed on the stacked branch `AIML-501-raise-coverage-tests` (#174) rather than here, because this PR has #174 stacked on top of it and changing this branch would have forced a restack. The changes flow down when the stack merges.

| # | Finding | Fix | Commit |
|---|---------|-----|--------|
| 1 | `check-test` ran the stdio-app suite twice | New `test-coverage` target does tests and floors in one gradle invocation, `check-test` calls it | 66ae436 |
| 2 | Per-counter NaN gap in the gate | `doFirst` now asserts each enforced counter measured something, not just the bundle | e81b11e |
| 3 | `always()` kept cancelled jobs alive | `!cancelled()` on both coverage steps | 697b1ed |
| 4 | Duplicated "No coverage minimums declared" | Extracted to `ext.requireCoverageMinimums` | e81b11e |
| 5 | Redundant `test` on the CI command line | Dropped | 697b1ed |
| 6 | `${project.name}` interpolated at execution time | Captured at configuration time as `moduleName` | e81b11e |
| 7 | Two near-identical `mustRunAfter` lines | Combined via `collectMany` | e81b11e |

A few notes where I did something other than the obvious.

**On #2**, I closed the gap rather than just commenting on it, since you were right that it was the one part of that guard not reasoned all the way through. The verification task now reads the JaCoCo XML in `doFirst` and fails if any enforced counter has a zero total. Mutation tested by adding a counter type JaCoCo never emits, which failed the build with the intended message.

That fix needed an XML parser, and duplicating the XXE hardening block would have been worse than the duplication you already flagged, so the `XmlSlurper` setup also moved into a shared `ext.jacocoXmlParser` closure alongside the #4 extraction.

**On #5**, I checked this rather than trusting the transitive dependency on faith. Ran the exact CI command from a clean tree and confirmed both module `test` tasks still execute.

**On #1**, worth noting it had quietly got worse than your review described. #174 added a fourth invocation, `buildsrc-check`, to the same chain.

Local verification is green. 1025 tests, 0 failures, both floors met, changed-file gate clean.

Separately, while verifying #5 I found a pre-existing flake in #174's own new tests, `SDKExtensionResponseHandlingTest.ApplicationsDebugLogging`, failing about one run in six. It was an SLF4J `SubstituteLogger` cast racing backend initialisation under parallel class execution, not anything to do with your findings. Fixed on the same branch and confirmed over 35 consecutive clean runs.

PR #174, opened 2026-07-27
AIML-501 Gate changed files at 85% coverage and raise module floors

ChrisEdwards:
**Jira:** [AIML-501](https://contrast.atlassian.net/browse/AIML-501)

## Summary
PR #173 added JaCoCo module floors set just under the measured baseline, which stops regression but does nothing to hold new code to a standard. This branch adds the second half, a per-changed-file gate at 85 percent line coverage enforced from CI and a pre-push hook, and then raises real coverage far enough that the new bar is met everywhere it applies. Core moved from 92.2 to 97.6 percent line and 82.5 to 89.1 percent branch. Stdio-app moved from 75.3 to 97.8 percent line and 69.2 to 96.2 percent branch.

Module floors were raised, but deliberately left a couple of points under the new measurements rather than flush against them. See the headroom note below.

## Why This Change Exists
The module floors from #173 are aggregate and deliberately low, so a large well-covered module absorbs an entirely untested new file without the build noticing. AIML-501 names AIML-934 in `aiml-services` as prior art for exactly this, a changed-file gate that holds new work to 85 percent regardless of what the module average happens to be.

The ticket also names AIML-1170, the defect where `aiml-services` sets its in-repo JaCoCo minimum to 0.0 and enforces the real number only through an external pinned action plus a skippable hook. That repo's gate is therefore fail-open. The instruction on this ticket was to not repeat it, so every number here is enforced by in-repo Gradle code, `verifyCoverageMinimums` fails the build if any floor is weakened, and the hook is a convenience layer over the same task CI runs rather than the only thing standing between a regression and `main`.

Raising coverage was the other half of the job. A gate set at 85 percent is not credible while two modules sit below it, and stdio-app at 75.3 percent line was well under.

## The Big Picture

```mermaid
flowchart LR
    subgraph analysis["buildSrc (plain Java, no Gradle types)"]
        CJF["ChangedJavaFiles<br/>git diff to file list"]
        JCF["JacocoChangedFileCoverage<br/>XML to per-file ratio"]
    end

    HOOK["pre-push hook"] --> TASK
    CI["CI, all pull requests"] --> TASK
    MAKE["make coverage-changed"] --> TASK

    TASK["jacocoChangedFileCoverageVerification"] --> CJF
    TASK --> JCF
    JCF --> REPORT["jacocoTestReport.xml"]

    CHECK["gradle check"] --> FLOORS["jacocoTestCoverageVerification<br/>per-module floors"]
    CHECK --> MINS["verifyCoverageMinimums<br/>floors cannot be weakened"]
    FLOORS --> REPORT

    BSCHECK["gradle -p buildSrc check"] --> BSOWN["buildSrc checkstyle, spotless,<br/>tests and its own floor"]
```

Key decisions:

- **The changed-file gate is deliberately not wired into `check`.** With no base ref it falls back to measuring the working tree, which is clean on a fresh CI checkout, so it would pass vacuously there. Locally it would block iteration on half-written code. It runs from CI on pull requests, the pre-push hook, and `make coverage-changed` instead. The module floors stay on `check`, where a fixed threshold is meaningful.
- **The gate fails closed, with a three-way outcome.** A changed file absent from the JaCoCo report fails, unless its source path is listed in `coverageExcludedClassFiles`. A missing or empty report fails. A file that is present in the report but carries no `LINE` counter has nothing countable in it, so it passes and is named in the output. That last case is not an edge case, it covers 49 of the 128 `contrast-mcp-core` sources, mostly interfaces and types whose members are entirely Lombok-generated.
- **Module floors keep headroom on purpose.** Set flush against the measurement, a floor has none. At 0.89 with core's 622 of 698 branches, one further uncovered branch anywhere gives 0.88984 and reddens `main` on an unrelated PR. That turns a regression guard into a tax on every commit and pushes people toward filler tests, which is the behaviour this change exists to remove. Floors are 0.95 line for both modules, 0.87 branch for core and 0.94 for stdio-app.
- **Analysis logic lives in `buildSrc` as plain Java with no Gradle types.** That makes it directly unit testable, which a Gradle task closure is not. It has 34 tests of its own, and `buildSrc` carries the same checkstyle and spotless configuration as the published modules plus its own JaCoCo floor, because the changed-file gate only inspects each module's `src/main/java` and so cannot measure the code that implements it.
- **`buildSrc` loads the root `gradle.properties` explicitly.** Gradle treats `buildSrc` as a separate build rooted at `buildSrc/` and does not apply the repository's `gradle.properties` to it. Duplicating the pinned versions there would create a second source of truth that silently drifts from the SECURITY.md soak policy, so the file is read directly instead, and a missing key fails at that point rather than later as a `:null` coordinate.
- **Property names match `aiml-services`** (`staticAnalysisChangedBase` and `staticAnalysisChangedHead`, plus `jacocoChangedBase` and `jacocoChangedHead`), so the hook ported across without rewriting its contract.
- **The hook measures only the ref whose commit is the checked-out `HEAD`.** Coverage comes from compiling and testing the working tree, so no other pushed ref can be scored honestly. Other refs are named and reported as not measured, and CI gates them. Base resolution is old remote tip, then the pushed ref's own upstream, then `$remote/main`, then `origin/main`. The upstream step is what keeps a stacked branch measured against its parent rather than against `main`, which would re-measure the parent's files as if they were new.

### 1. Changed-file analysis (Focus)
The whole gate reduces to two pure functions. Worth reading first because everything else just calls them.
- `buildSrc/src/main/java/com/contrast/labs/build/JacocoChangedFileCoverage.java` (new): parses the report, computes a per-file covered ratio, and sorts each changed file into measured, nothing-to-measure, or never-measured. Also maps `coverageExcludedClassFiles` onto the source paths they hide, so a deliberate exclusion is not mistaken for a file the report missed.
- `buildSrc/src/main/java/com/contrast/labs/build/ChangedJavaFiles.java` (new): resolves the changed set from git. An omitted base means working tree mode including untracked files, while an explicitly blank base property fails; a non-blank base uses a three-dot `base...head` range. Paths the diff names that the working tree no longer holds are returned separately rather than dropped.
- `buildSrc/build.gradle` (new) and `gradle.properties`: toolchain, the checkstyle, spotless and JaCoCo wiring, and the pins the loader reads.

### 2. Gradle wiring and floors (Focus)
- `build.gradle`: registers `jacocoChangedFileCoverageVerification` per subproject, adds `changedFileCoverageMinimum` at 0.85, adds `verifyCoverageMinimums` and `installGitHooks`, and sets the four module floors.

### 3. Enforcement paths (Skim)
Three entry points, one task. Nothing here contains logic beyond resolving a base ref.
- `.github/workflows/build.yml`: the changed-file step is pull-request only, `pull_request` carries no base-branch filter so stacked PRs are gated, `buildSrc` gets its own step because the main build never schedules its tests, and the checkout uses `fetch-depth: 0` because depth 1 fetches neither the PR base commit nor the merge base.
- `scripts/git-hooks/pre-push` (new): base resolution as described above, bypass with `SKIP_COVERAGE_HOOK=1`.
- `Makefile`, `scripts/git-hooks/README.md`, `CLAUDE.md`, `AGENTS.md`, `CONTRIBUTING.md`: `make coverage-changed`, `make buildsrc-check`, `make install-hooks`, and the docs for all three.

### 4. The tests carrying the most risk (Focus)
These three are where a wrong assertion would be least obvious.
- `.../tool/vulnerability/VulnerabilityMapperTest.java` (new): both target records are long runs of same-typed `String` fields, and the two mapper methods order their timestamps differently, so a transposed argument compiles clean and ships silently. Every field gets a distinct value and three distinct timestamps.
- `.../sdkextension/SDKExtensionResponseHandlingTest.java` (new): pagination, the debug-buffering branch, and the null-response defaults.
- `.../sdkextension/SDKHelperTest.java` (modified): caching and paging behaviour. This file also had a latent race, covered under Testing.

### 5. Remaining new tests (Skim)
- `.../result/RunBookEnumTest.java`, `.../data/server/ServersResponseEnvelopeTest.java`, `.../data/application/ApplicationTest.java`, `.../data/LibraryVulnerabilityExtendedTest.java`, `.../data/adr/AttacksResponseTest.java`, `.../tool/base/CursorExecutionResultTest.java`, `.../config/SDKExtensionFactoryTest.java` (all new).

<details>
<summary>buildSrc self-tests (2 files, 34 tests)</summary>

`ChangedJavaFilesTest` builds a real temporary git repository rather than mocking git, and covers non-ASCII filenames, renames, a file deleted from the working tree, and an unresolvable base ref. `JacocoChangedFileCoverageTest` carries the JaCoCo DOCTYPE in its fixture deliberately, since that declaration is what makes the default parser configuration reject the file.

</details>

## Testing
`make check-test` passes with 1,026 unit tests, zero failures and zero skips, plus 34 tests in `buildSrc`. buildSrc coverage is 93.7 percent line and 98.5 percent branch against its own 85 percent line and branch floors.

Every production behaviour claimed above was mutation tested rather than assumed. 39 mutations were applied across the classes under test and 38 were killed outright. The one survivor is worth calling out because it is a real finding rather than a gap. Removing the `typeID != null` guard in `RunBookEnum` on its own changes nothing, since `equalsIgnoreCase(null)` returns false rather than throwing. That guard only earns its place if someone also flips the receiver to `typeID.equalsIgnoreCase(...)`, and the null test does kill that combination. So the test pins a real contract, not a redundant line.

The gate itself was driven end to end against three files chosen to hit each outcome. A Lombok-only DTO and the excluded `McpContrastApplication` both passed as nothing-to-measure and were named in the output. `LocalSdkPaginatedTool` at 0 percent coverage failed at 85 percent. `verifyCoverageMinimums` was verified with `-PcoverageStandardMinimum=0.0` and fails against a literal 0.85 invariant, which is the AIML-1170 shape the ticket forbids. The pre-push hook was exercised for a non-HEAD ref, a delete-only push, and the `SKIP_COVERAGE_HOOK=1` bypass.

One latent defect was found and fixed along the way. `SDKHelperTest` mutated `SDKHelper`'s static `environment` field while this repo runs test methods concurrently by default, so the class now carries `@Execution(ExecutionMode.SAME_THREAD)`.

Several SDK model classes have no setters and are only ever populated by Gson from a TeamServer payload. Tests for those use wire-format JSON fixtures instead of hand-built objects, which both compiles and exercises the path the code actually takes in production.

## Review Round
A multi-model review of the first version of this branch found five defects that weakened the gate, four of them against the ticket's own fail-closed instruction. All are fixed in `a404aad`. The substantive ones:

- The gate passed silently on any file JaCoCo could not measure, which was 49 of 128 core sources rather than the single excluded class the docs claimed. It now separates never-measured from nothing-to-measure and fails on the former.
- Skip output went to `LIFECYCLE`, which the hook's `-q` suppresses, and the hook then discarded its captured output on success. Neither is true now.
- `buildSrc` sat outside every check in the repository. `:buildSrc:test` was in no build path, and it had no checkstyle, spotless or coverage.
- Nothing read the floors back, so setting one to 0.0 left the build green.
- The `pull_request` base-branch filter meant CI never ran on a stacked PR, so this PR received no gate run at all before now.

A later GOAT review reproduced three remaining gaps, fixed in `3e33aa9` and `9d422f9`: the documented floor compared a property-derived value to itself and could be overridden to zero; an explicitly blank changed-base property passed without measuring files; and `buildSrc` verification could skip green without execution data. The same pass added a mutation-sensitive proxy wiring test, made zero-file runs visible, and corrected the per-file-floor wording.

Smaller fixes: `core.quotepath` was silently dropping non-ASCII filenames from the gate, `--end-of-options` now stops a leading-dash base ref being read as a git option, the hook resolved `HEAD`'s upstream rather than the pushed ref's and had an unguarded `merge-base` that aborted the push, and `installGitHooks` wrote to a hard-coded `.git/hooks` that breaks in a worktree and silently clobbered any hook already there.

- **`LocalSdkPaginatedTool` sits below the new gate and is left uncovered on purpose.** It is four lines, every method a single `return factory.getX()` delegation, with no subclass anywhere in the repo. A test would mock the factory and assert the getter returned what the mock was told to return, which proves nothing. The consequence is real though, editing it will fail the gate, and at that point the options are to cover it or to exclude it by name. Note that `LocalSdkSingleTool`, its near-duplicate, is *not* in the same position. It is 75 percent covered and `GetSastResultsTool` extends it. The two are byte-for-byte identical apart from the class declaration, which is the better thing to fix and is out of scope here.
- **The pre-push hook is opt-in.** It needs `make install-hooks`, and CI is the backstop for anyone who has not run it. Installing it will displace any other `pre-push` hook, including the beads one. The task now backs up what it replaces and says so, but it cannot merge the two.
- **The Gradle task wiring has no functional test.** The pure analysis logic is covered in `buildSrc` and the configured minimums are pinned by `verifyCoverageMinimums`, but the `doLast`, `onlyIf` and provider-based `dependsOn` are verified by hand only. A Gradle TestKit test would need its own fixture build and would add minutes to CI, so it is deliberately deferred.
- **`RunBookEnum` has no caller in this repository**, and never has had one. It is not simply dead, though. `contrast-mcp-core` applies `maven-publish`, so the class is published API surface, and commit `00fcf9b` moved it into core for the `aiml-services` consumer. Whether it should be wired up here or left as published surface is a separate decision.
- **The changed-file gate is not configuration-cache compatible.** It shells out to git while resolving providers. The provider is memoized so it runs git once per module instead of four times, but a `ValueSource` is the real fix if the cache is ever enabled.

[AIML-501]: https://contrast.atlassian.net/browse/AIML-501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/SDKHelperTest.java
```diff
@@ -254,4 +274,388 @@ void getLibraryPage_should_return_single_page_with_count() throws IOException {
...
+
+    var sdk =
+        SDKHelper.getSDK(
+            "https://custom.example.com", "apiKey", "serviceKey", "user", null, null, "https");
+
+    assertThat(sdk.getRestApiURL()).isEqualTo("https://custom.example.com/Contrast/api");
+  }
+
+  @Test
+  void getSDK_should_build_an_sdk_when_a_proxy_host_and_port_are_supplied() {
```
Alex-Contrast: asked claude to look at the tests to make sure they're good. it pointed at these 2 as coverage oriented.
```
one real problem

  SDKHelperTest.java:567 and :578 are coverage-shaped, not behavioral. Both assert only sdk.getRestApiURL(), which the proxy branch never touches, so
  the assertion is identical to the no-proxy test above it. I confirmed you can delete the entire withProxy call and both stay green. The second one
  is worse because its name promises the port defaults to DEFAULT_HTTP_PROXY_PORT (80) and nothing anywhere asserts 80.

  Whether that is fixable depends on whether ContrastSDK exposes the proxy. I could not find the SDK source at ../contrast-sdk-java to check. If there
  is no accessor, the honest move is to delete these two tests and let the line go uncovered rather than keep a test that asserts nothing. The
  sibling test at :589 (invalid port throws NumberFormatException) is genuine and does pin the parse behavior, so most of the branch is legitimately
  covered anyway.

  I'd go with the extraction. It is a small change to SDKHelper, it turns two dead tests into real ones, and it does not touch the SDK boundary.

  Extract the port resolution. The real logic in that branch is the port decision, not the withProxy call. Pull it into something like
  SDKHelper.resolveProxyPort(String) and the behavior becomes directly assertable. Blank or null gives 80, "8080" gives 8080, "http" throws.
  getSDK_should_default_the_proxy_port_when_only_a_proxy_host_is_supplied then actually asserts 80, which its name has been promising all along. The
  existing getSDK_should_reject_a_proxy_port_that_is_not_a_number moves across too and stops being the only honest test of the three. Coverage of the
  branch is preserved, so no floor risk.
```

dougj-contrast (commented):
## Java PR Review: Gate changed files at 85% coverage and raise module floors

Independent review pass (Claude + Codex CLI + Copilot CLI) against this diff. Unusually rigorous, self-aware work overall — most of what a first pass would normally catch, this PR already caught and documented itself.

### Important
- **Rounding-before-comparison lets a file just under 85% pass the gate** (`buildSrc/.../JacocoChangedFileCoverage.java`, `FileCoverage.coveredRatio()`) — `coveredRatio()` rounds to 4 decimal places (`HALF_UP`) *before* `Result.failuresBelow()` compares it against the minimum. Concretely verified: 861/1013 covered lines = 84.9951% (genuinely below 85%), but rounds to `0.8500`, which is `>= 0.85` and passes. This affects **both CI and the local hook**, since they share this exact task. Suggested fix: compare the exact fraction (cross-multiply `coveredLines * 10000` against `minimumRatio * totalLines * 10000` via `BigDecimal`) and keep the 4-dp rounding for the display string only.
- **Pre-push hook can score a different tree than what's being pushed** (`scripts/git-hooks/pre-push`) — the hook diffs *committed* refs (`base...head`) to decide which files changed, but Gradle's test run (which produces the coverage numbers) measures whatever's actually on disk at that moment. Uncommitted changes at push time — extra test coverage not yet committed, or an uncommitted deletion of a changed file — make the hook's read diverge from the commit actually being pushed. CI is unaffected (clean checkout every time); this is a local-hook-only gap, and the hook is opt-in with CI as the documented backstop, so severity is bounded.
- **Two coverage-shaped, not behavioral, tests in `SDKHelperTest.java`** (`getSDK_should_build_an_sdk_when_a_proxy_host_and_port_are_supplied` and `getSDK_should_default_the_proxy_port_when_only_a_proxy_host_is_supplied`) — both only assert `sdk.getRestApiURL()`, which the proxy branch never touches; deleting the entire `withProxy` call leaves both green, and the second test's name explicitly promises the port defaults to 80 with nothing asserting that. Alex already flagged this — confirming it's still present in the current HEAD. Alex's suggested fix (extract `SDKHelper.resolveProxyPort(String)` so the port decision is directly assertable) looks right and doesn't touch the SDK boundary.
- **Worth verifying, not confirmed**: `find_changed_base`'s fallback order in `pre-push` tries `$upstream_ref@{upstream}` before falling back to `origin/main`. On the very first push of a brand-new stacked branch, if the local branch has no upstream configured yet (git typically sets that *after* a successful push, not before), this would silently fall through to gating against `main` — which the PR's own text says is exactly the scenario it's trying to avoid ("would re-measure the parent's files as if they were new"). Worth a quick check on whether the stacked-branch tooling here pre-sets the upstream before the first push.

### Code smells (optional polish)
- **`buildSrc`'s own coverage floor sits below the standard it enforces everywhere else** — hardcoded `LINE 0.85 / BRANCH 0.80` directly in `buildSrc/build.gradle`'s violation rule, not sourced from `coverageMinimums`, and structurally invisible to the new `verifyCoverageMinimums` meta-guard (buildSrc is a separate Gradle build, not one of `subprojects`). `BRANCH 0.80` is already below the "documented 0.85 standard" the rest of this PR is built around enforcing, with nothing stopping it drifting further.
- `changedFileCoverageMinimum` and `documentedCoverageFloor` are two separate constants that happen to both be `0.85` today, with no comment on whether that's coincidence or a rule.
- Minor, low priority: the hook's own changed-set check doesn't set `core.quotepath=false` (benign — won't cause a wrong skip); `find_changed_base`'s unrelated-history edge case skips the ref rather than falling through further (rare, CI backstops it); `gradle.properties` parsed twice in `buildSrc/build.gradle` (unavoidable given buildscript ordering); `changedFilesHolder`'s memoization is a benign, not-technically-thread-safe cache (idempotent result, harmless in practice).

### Good
- XXE hardening in the new DOM-based `JacocoChangedFileCoverage.parse` is thorough, matching and in places exceeding PR #173's `XmlSlurper` approach.
- `verifyCoverageMinimums` directly targets the AIML-1170 defect class (a floor silently set to 0.0) and is wired into both `check` and `jacocoTestCoverageVerification`.
- `ChangedJavaFiles`'s git invocation is careful: `core.quotepath=false` + NUL-delimited output for non-ASCII filenames, `--end-of-options` against a dash-leading ref, virtual-thread stderr draining to avoid a pipe deadlock.
- `installGitHooks` resolves the real hooks directory via `git rev-parse --git-path hooks` (worktree/`core.hooksPath`-safe), backs up rather than silently clobbering a pre-existing hook (relevant since `beads` already installs its own `pre-push`), and sets safe `0700` permissions on the installed script.
- CI hardening: unfiltered `pull_request` trigger (correctly reasoned to cover stacked PRs), `fetch-depth: 0`, `persist-credentials: false` on a job running PR-authored build code, base SHA passed via `env:` rather than interpolated directly into the shell.
- Fail-closed design throughout (missing/empty report, unreported-and-not-excluded, absent-from-working-tree all throw).

Not requesting changes yet — flagging the rounding bug as the one item worth fixing before merge since it's a real gap in the mechanism this PR exists to add, and the fix is small.

dougj-contrast (approved):

contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/sdkextension/SDKHelperTest.java
ChrisEdwards: Addressed in 0d52f33. I extracted `SDKHelper.resolveProxyPort(String)`, wired `getSDK` through it, and replaced the coverage-shaped URL assertions with direct behavioral assertions: `"8080"` resolves to 8080, blank/null resolves to 80, and `"http"` throws `NumberFormatException`.

Verified with the full `SDKHelperTest` and the repository test/coverage suite.

ChrisEdwards:
@dougj-contrast Thanks for the thorough pass. I addressed the actionable findings from [this review](https://github.com/Contrast-Security-OSS/mcp-contrast/pull/174#pullrequestreview-4788269239):

* `3f7c31f` compares the exact covered-line fraction against the minimum and only rounds for display. The regression test uses the reported 861/1013 case and confirms it fails at 85% even though its display ratio rounds to `0.8500`.
* `7498ebf` prevents the pre-push gate from scoring coverage-relevant working-tree changes against a different pushed commit. Follow-up `b3aaf68` narrows this correctly: dirty source/build/resource files block, while unrelated changes such as Markdown are allowed.
* `0d52f33` extracts `SDKHelper.resolveProxyPort` and replaces the two coverage-shaped proxy tests with direct assertions for explicit, default, and invalid ports.
* I reproduced the first-push concern: Git does not install the `push -u` upstream until the push succeeds. `7498ebf` therefore adds `COVERAGE_BASE_REF=origin/<parent>` for the first push of a stacked branch and documents it.
* `76b1238` makes `coverageStandardMinimum=0.85` the shared source for the changed-file gate, the meta-guard, and both `buildSrc` counters, raising the `buildSrc` branch floor from 80% to 85%.
* `7498ebf` also adds `core.quotepath=false` to the hook diff and lets unrelated-history failures continue through the remaining base candidates.

I left two explicitly benign items unchanged: `buildSrc` must read `gradle.properties` once during the early `buildscript` classpath phase and again in the normal script phase, and `changedFilesHolder` caches an idempotent provider result within one Gradle configuration lifecycle. Neither creates a correctness gap worth extra machinery.

Verification: `make check-test` and `make buildsrc-check` pass; 1,025 tests, core 97.6% line / 89.1% branch, stdio-app 96.9% line / 95.2% branch. The updated hook was also exercised both ways: dirty Markdown was allowed and an untracked Java file was rejected.

ChrisEdwards:
Follow-up to my review response: `9eed859` refines the dirty-worktree behavior after maintainer discussion. The hook no longer skips coverage or blocks merely because source/build/resource files are dirty. It runs the coverage task against the current working tree, and coverage below 85% remains a hard failure that blocks the push. The dirty-tree message only explains what was measured; pull-request CI still verifies the exact commit from a clean checkout.

dougj-contrast (approved):

ChrisEdwards:
Follow-up on the GOAT coverage-gate review:

* `3e33aa9` makes the documented 85% standard a literal invariant, rejects explicitly blank changed-base properties, reports zero-file runs, and makes `buildSrc` coverage verification fail when execution data, class files, the XML report, or enforced counters are missing.
* `9d422f9` adds a behavioral proxy-wiring test that fails if `builder.withProxy(proxy)` is removed, clarifies that 85% is a separate per-file constraint rather than numerically above the module floors, documents `make buildsrc-check` in AGENTS.md, and refreshes the post-fix coverage figures.

Verification:

* `make buildsrc-check` — passed
* `make check-test` — passed; 1,026 tests, zero failures/skips; core 97.6% line / 89.1% branch, stdio-app 97.8% line / 96.2% branch
* `make coverage-changed BASE=origin/main` — passed
* `./gradlew verifyCoverageMinimums -PcoverageStandardMinimum=0.0` — failed as expected against the literal 0.85 invariant
* `./gradlew -q jacocoChangedFileCoverageVerification -PjacocoChangedBase=` — failed as expected because the supplied base was blank
* `./gradlew -p buildSrc clean jacocoTestCoverageVerification -x test` — failed as expected because coverage was not measured
* Proxy mutation check — removing `builder.withProxy(proxy)` made the new test fail; restoring it made the test pass

The remaining generated-code, cross-class isolation, and advisory pre-push observations were left unchanged for the reasons documented in the triage. This GOAT report was not a GitHub code-review thread, so there is no thread to resolve.

PR #175, opened 2026-07-28
AIML-1245 Auto-approve and auto-merge minor and patch Dependabot PRs

ChrisEdwards:
**Jira:** [AIML-1245](https://contrast.atlassian.net/browse/AIML-1245)

## Summary

Dependabot's minor and patch version bumps now approve themselves and merge once CI passes, so routine dependency updates stop needing a human. Dependabot PRs are also ungrouped, which is what makes the automation useful rather than decorative.

- Minor and patch bumps are approved and auto-merged. Majors are left alone.
- `build` must pass before anything merges, so nothing lands with red required CI.
- One PR per bump instead of one PR per group.

## Why This Change Exists

Dependency bumps were merged by hand. That is a steady tax on someone's attention for changes that are almost always safe, and it means bumps sit unmerged for days.

Grouping made the automation ineffective on its own. `dependabot/fetch-metadata` reports a single update type per PR, taken from the highest semver change in the group, so one major bump in the production group made every minor and patch bump beside it ineligible. Ungrouping is load-bearing here, not a side tidy-up.

Approving from a workflow is safe only with a required CI gate. Before this work, the ruleset required one approving review and no status check; because this workflow supplies the approval itself, a bump could otherwise land with red CI. Repository auto-merge is enabled and `build` is now required.

`strict_required_status_checks_policy` is intentionally disabled, matching `adr-services`. Every PR must pass `build`, but it does not need to rebase and rerun after another PR advances `main`. This lets an ungrouped weekly batch drain without serialized rebases. The tradeoff is that independently green bumps may merge without first being tested together; the `main` push build is the combined-state backstop.

## Approach

Copied from the shared composite action at `Contrast-Security-Inc/common-github-actions/dependabot-auto-approve-merge`, upstream commit `fc43269` (ADR-738), then vendored. It cannot be referenced across the org boundary because that repository is internal to the Inc org while this repository is public in Contrast-Security-OSS.

Adapted rather than copied verbatim:

- **`secrets.GITHUB_TOKEN` with explicit `contents: write` and `pull-requests: write`**, rather than the two shared org PATs upstream uses. GitHub documents that explicit workflow permissions can raise a Dependabot-triggered run's initially read-only token.
- **`--merge` rather than `--squash`.** Squash is disabled here and the PR Approvals ruleset permits only merge and rebase.
- **Actions pinned to full commit SHAs**, including `dependabot/fetch-metadata` v3 and `step-security/harden-runner`.
- **No `actions/checkout`.** The workflow does not read or execute the pull request working tree.
- **Eligibility is one exact, fail-closed check**, so an empty or unrecognized update type remains for a person.

A GitHub App token is the intended end state, tracked as bead `mcp-vc50`. John Ament is moving the shared action off its static PAT onto an App, and the OSS org should follow.

### Workflow

- `.github/workflows/dependabot-auto-merge.yml` (new) — runs on Dependabot pull-request events, reads the update type, approves and enables auto-merge for minor and patch updates, and logs/skips everything else.

### Dependabot config

- `.github/dependabot.yml` — removes `groups:` from Gradle, GitHub Actions, and Docker. Raises `open-pull-requests-limit` to 15 for Gradle and 10 for GitHub Actions so the default of five does not throttle an ungrouped weekly run. Retains the seven-day cooldown on all three ecosystems, preserving the ENTSEC-1742 soak window.

## Testing

`actionlint` is clean on the new workflow and across all repository workflows. `make check` passes.

The eligibility `case` was extracted and run against seven update types. `version-update:semver-minor` and `version-update:semver-patch` are eligible. Major, empty, unknown, malformed-minor, and extra-suffix inputs are ineligible, confirming that the workflow fails closed.

Live repository settings confirm:

- auto-merge enabled;
- GitHub Actions may approve pull requests;
- squash disabled;
- `build` required; and
- strict up-to-date enforcement disabled by policy decision.

## Review Findings Still Open

- **Actor versus author.** The job currently checks `github.actor == 'dependabot[bot]'`, which tests who initiated an event rather than who authored the PR. A Dependabot PR reopened or synchronized by a maintainer would be skipped. If author identity is intended, use `github.event.pull_request.user.login == 'dependabot[bot]'`.
- **PR-context action execution.** Omitting checkout keeps PR working-tree code off the runner, but a `pull_request` workflow still runs the PR-context workflow definition. A Dependabot update to an action pin in this workflow can therefore execute the proposed action version with the job's write token. This must be resolved or explicitly accepted, and the GitHub App follow-up must mint its persistent credential only from a trusted default-branch workflow definition.

## Live Rollout Gap

The real path cannot be exercised from this feature PR. The workflow only acts on Dependabot events, and future Dependabot branches must contain or be evaluated against the merged workflow. The open Dependabot PRs #164 and #155 are major bumps, so deliberately enabling auto-merge on them would be unsafe.

On the first eligible minor or patch update, verify:

1. The job runs and records the expected approval identity.
2. Auto-merge waits for `build` to pass.
3. Major and unknown update types remain untouched.
4. A `wiz-scan` push run exists on the merge commit SHA.

- **The post-merge scan may go quiet.** Events attributed to `GITHUB_TOKEN` normally do not trigger further workflow runs. If that applies to the eventual GitHub-performed auto-merge, `wiz-scan` will not run on the merge commit; because it also cannot authenticate on the Dependabot PR, the bump would be unscanned. Verify the first auto-merge. The fix is the trusted GitHub App design in bead `mcp-vc50`, currently blocked on IT provisioning.
- **Green PRs are not retested together before merge.** Disabling strict up-to-date enforcement lets a batch drain quickly, matching `adr-services`, but two individually green dependency changes can interact. The `main` push build is the combined-state backstop.
- **`wiz-scan` is already red on Dependabot PRs.** Its credentials are in the Actions secret store, which Dependabot-triggered runs cannot read. It is not required and does not block a merge. This is pre-existing and tracked as bead `mcp-f946`.

[AIML-1245]: https://contrast.atlassian.net/browse/AIML-1245

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,20 +1,16 @@
 version: 2
+# Deliberately ungrouped. dependabot/fetch-metadata reports a single update-type
+# for a grouped PR, taken from the highest semver change in the group, so one
+# major bump made every minor and patch bump in that group ineligible for the
+# auto-merge workflow. One PR per bump means each is judged on its own change.
+# open-pull-requests-limit is raised past the default of 5 so ungrouping does not
+# throttle a weekly run to five bumps.
 updates:
   - package-ecosystem: "gradle"
     directory: "/"
     schedule:
       interval: "weekly"
-    groups:
-      dev-dependencies:
-        applies-to: version-updates
-        patterns:
-          - "*"
-        dependency-type: development
-      production-dependencies:
-        applies-to: version-updates
-        patterns:
-          - "*"
-        dependency-type: production
+    open-pull-requests-limit: 15
     cooldown:
       default-days: 7
 
@@ -22,17 +18,7 @@ updates:
     directory: /
     schedule:
       interval: weekly
-    groups:
-      dev-dependencies:
-          applies-to: version-updates
-          patterns:
-              - '*'
-          dependency-type: development
-      production-dependencies:
-          applies-to: version-updates
-          patterns:
-              - '*'
-          dependency-type: production
+    open-pull-requests-limit: 10
     cooldown:
       default-days: 7
       include:
@@ -44,9 +30,5 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-    groups:
-      docker-dependencies:
-        patterns:
-          - "*"
     cooldown:
       default-days: 7
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -15,6 +15,11 @@ name: Dependabot Auto-Approve and Merge
 # to GITHUB_TOKEN does not trigger further workflow runs, which may stop the post-merge
 # wiz-scan firing on an auto-merged bump. Migration is one create-github-app-token step
 # plus swapping the token references below. Tracked in mcp-vc50.
+#
+# The self-modification guard job below has no upstream counterpart, and does not need
+# one: upstream keeps its action pins in common-github-actions, so a consuming repo's
+# Dependabot cannot alter what runs inside the approving job. Vendoring put the pins and
+# the auto-merge in the same repository, which is what creates the loop the guard closes.
 on:
   pull_request:
     types: [opened, synchronize, reopened]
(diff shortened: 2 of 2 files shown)


Alex-Contrast (changes requested):
**Requesting changes on one item. The rest is verified and two notes.**

Verified against live settings and upstream. Both action pins resolve to their commented tags, `build` is the only required check and passes on Dependabot PRs without secrets, `--merge` is correct since squash is disabled and the ruleset allows only merge and rebase, actionlint is clean, and the eligibility case is fail-closed across ten inputs including glob characters and case variants. Every adaptation from `fc43269` is real. The description is accurate throughout.

**Blocking, self-modification.** `github-actions` is enabled at `directory: /`, so Dependabot will propose minor and patch bumps to this file's own action pins. The `pull_request` run uses the PR-context definition, so the proposed action version executes with `contents: write` and `pull-requests: write` and then approves and merges itself. Your description says this must be resolved or accepted, and it is reachable rather than theoretical. Guard needs no checkout.

```yaml
- name: Leave self-modifying updates for a human
  env:
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    PR_URL: ${{ github.event.pull_request.html_url }}
  run: |
    if gh pr diff --name-only "$PR_URL" | grep -qx '.github/workflows/dependabot-auto-merge.yml'; then
      echo "PR modifies the auto-merge workflow itself, needs human review."
      exit 1
    fi
```

**Not blocking, but new.** The approval and armed auto-merge survive a later push. `dismiss_stale_reviews_on_push` and `require_last_push_approval` are both false, Dependabot branches are in-repo and unprotected, and GitHub only disarms auto-merge for pushes from users *without* write access. So a write-access holder can push to an approved Dependabot branch and land it on `main` with only the bot's approval. Fix is `require_last_push_approval: true` on the ruleset, not in this file.

**Not blocking, a warning.** Do not make your actor-versus-author change as written. `fetch-metadata` validates `commits[0]` only, so switching to `pull_request.user.login` would let the job approve a branch whose tip is human-authored. Require both conditions instead.

JacobMagesHaskinsContrast (approved):

ChrisEdwards:
Thanks Alex, the self-modification loop is real and I have taken it. All three items below, addressed in 898d7a6, with one adjustment to your fix and one item that belongs outside this PR.

### Blocking item, fixed, but placed differently to your snippet

The guard now runs as its own read-only `guard` job ahead of `auto-merge`, rather than as a step inside it.

Placement turned out to be the part carrying the value. Your snippet sits after `harden-runner` and `fetch-metadata`, so on a PR bumping either pin the proposed version has already executed by the time the guard runs. And it does not need the approve and merge steps to finish the job. That job holds `pull-requests: write`, and I confirmed `can_approve_pull_request_reviews` is `true` on this repo, so malicious code in step 1 can open, approve and merge inline without touching anything below it. The dangerous grant is the `permissions:` block, not the steps under it. Gating before any `uses:` step runs means an untrusted action version never executes at all, rather than executing and then being refused a merge.

A separate job rather than just an earlier step, because a step-level gate needs a matching `if:` on `harden-runner`, `fetch-metadata` and `Check eligibility` to be equivalent. Three conditions that all have to stay correct are a worse thing to maintain than one `needs:` edge.

Verified `gh pr diff --name-only` resolves from the PR URL with no checkout and no git remote present, returning true for this PR and false for 173 and 174. `gh` is preinstalled on the runner, so the guard needs no action of its own. actionlint is clean and it ran shellcheck over the `run` blocks.

I also considered ignoring `dependabot/fetch-metadata` in `.github/dependabot.yml`, which would give us back the structural property upstream has for free. I left it out. The guard covers both pins whereas an ignore would cover one, and I would rather not quietly stop hearing about `fetch-metadata` releases. Say the word if you want both.

### Actor versus author, taken as advised

Both conditions now, on both jobs. Your reasoning is recorded in a comment on the file so it does not get simplified back to a single check later. The duplication onto `auto-merge` is deliberate, since that is the job holding write permissions and it should not depend on `needs: guard` surviving a future edit in order to stay gated.

### require_last_push_approval, agreed, and deliberately not done here

Confirmed your reading. All three rulesets on the default branch have empty `bypass_actors`, and `require_last_push_approval` is false on both that carry a `pull_request` rule. Since this is a ruleset change rather than a file change, and it makes every contributor re-approve after any push, I have not flipped it unilaterally. Chris is deciding, and it would land as a settings change referenced from AIML-1245 rather than in this PR.

Worth adding that those rulesets do block direct pushes to `main`, including for the Actions token, so the approve and merge path really is the only way anything lands. That makes this item more load bearing than it first looks.

### One thing found while checking upstream, probably for team-cratus rather than this PR

Upstream has no guard and does not need one. Its pins live in `common-github-actions`, so a consuming repo's Dependabot cannot change what runs inside the approving job. Vendoring put the pins and the auto-merge in the same repository, and that is what created this loop. I noted it in the header comment so the divergence is not later mistaken for drift.

While confirming that, though, upstream floats `actions/checkout@v4` and `dependabot/fetch-metadata@v2` on major tags, and passes two static PATs into the job rather than a scoped `GITHUB_TOKEN`. A re-pointed tag is the version of this attack that actually happens in the wild, and roughly twelve Inc repos consume that action. Bigger exposure than this PR ever had, and nothing we can fix from here, but someone should tell them.

Alex-Contrast (approved):

PR #179, opened 2026-07-30
AIML-942 Rename MCP response notices and remediation hint

ChrisEdwards:
**Jira:** [AIML-942](https://contrast.atlassian.net/browse/AIML-942)

## ℹ️ _NOTE!_
_**This is a large PR, but almost entirely renames.**_

_I'm building a mechanism to reduce context in tool descriptions. Today, all guidance notes are embedded in every description, and customers are reporting token usage is too high (20k tokens, which I agree is excessive). The fix is to attach these notes dynamically as "Notices" only when specific conditions arise._

_The rename is the bulk of this diff. These messages were previously called "Warnings," which caused agents to treat them as problems when they aren't. Renaming to "Notices" fixes that. I underestimated how many files the rename would touch._

## Summary

Renames the MCP response envelope's `warnings` field to `notices` across all three response types and renames the `Vulnerability` record's `hint` field to `remediationHint`. This is the wire-format rename that MCP_STANDARDS v2 deferred behind a migration note, now shipped. Both renames change what AI agents see on the wire, aligning field names with what the data actually represents.

- Response envelope field `warnings` → `notices` on `PaginatedToolResponse`, `SingleToolResponse`, and `CursorToolResponse`
- Internal `WarningCollector` → `NoticeCollector` with method renames (`warn()` → `notice()`, etc.)
- `Vulnerability.hint` → `Vulnerability.remediationHint` for self-documenting field semantics

## Why This Change Exists

The parent PR (AIML-942 canonical description standard) introduced `notices` as the documented name for informational messages in MCP_STANDARDS v2, but kept the wire field as `warnings` with a migration note: "The wire field is named `warnings` until the AIML-942 envelope rename ships." This PR ships that rename.

The old name `warnings` primed AI agents to treat informational teaching notes as problems and waste turns "fixing" normal behavior. `notices` better reflects the intent established in ADR 0001 (notices tier and point-of-use teaching).

The `hint` → `remediationHint` rename follows the same principle from MCP_STANDARDS: "Prefer renaming a misleading field over documenting it. A rename costs zero recurring tokens." Bundled here so the wire shape breaks only once. Broader renames (`environments` → `environmentsLatestInstance`, `httpRequest` → `httpRequestRedacted`) were considered and declined because the clarity gain didn't justify the breaking change. Those facts stay as standing description clauses instead. The MCP_STANDARDS example was updated to reference the shipped `remediationHint` rename rather than the declined `httpRequestRedacted`.

### 1. Wire format proof — **Focus**
Two new test files prove the wire shape changed correctly. `ResponseEnvelopeSerializationTest` serializes each response type to JSON and asserts `"notices"` is present while `"warnings"` is absent. `VulnerabilityTest` does the same for `remediationHint` vs `hint`.

- `contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/ResponseEnvelopeSerializationTest.java` (new) — serialization contract for all three envelope types
- `contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/result/VulnerabilityTest.java` (new) — serialization contract for Vulnerability record

### 2. Core rename: WarningCollector → NoticeCollector — **Focus**
The class rename and all method renames (`warn` → `notice`, `warnForEmptyResults` → `noticeForEmptyResults`, `hasEmptyResultsWarning` → `hasEmptyResultsNotice`). Behavior is unchanged.

- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/NoticeCollector.java` (renamed from `WarningCollector.java`)
- `contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/NoticeCollectorTest.java` (renamed from `WarningCollectorTest.java`)

### 3. Response envelope records — **Skim**
Field rename `warnings` → `notices` in the three response records and their factory methods. No logic change.

- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/PaginatedToolResponse.java`
- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/SingleToolResponse.java`
- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/CursorToolResponse.java`

### 4. Base tool classes and params — **Skim**
Updating `WarningCollector` references to `NoticeCollector` in base tool pipelines, and `warnings()` to `notices()` in params.

- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/PaginatedTool.java`
- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/SingleTool.java`
- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/CursorPaginatedTool.java`
- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/BaseToolParams.java`
- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/ToolParams.java`
- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/PaginationParams.java`
- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/CursorPaginationParams.java`

### 5. Vulnerability record — **Skim**
Single field rename `hint` → `remediationHint`.

- `contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/result/Vulnerability.java`

### 6. Standards doc — **Skim**
Removes the migration note from MCP_STANDARDS.md and updates the rename example to use the shipped `remediationHint` instead of the declined `httpRequestRedacted`.

- `MCP_STANDARDS.md`

<details>
<summary>Mechanical / low-risk changes (83 files)</summary>

Tool implementations updating `WarningCollector` → `NoticeCollector` import and `doExecute` signature (13 files), `warn()` → `notice()` / `warnIf()` → `noticeIf()` call sites (5 files), validation context `warnIf` → `noticeIf` and `warnings` → `notices` (7 files), and all corresponding test files updating assertions from `warnings()` to `notices()` and `warn()` to `notice()` (58 files). No logic changes in any of these files.

</details>

## Testing

- `ResponseEnvelopeSerializationTest` proves the wire format: all three response types serialize `notices` and do not emit `warnings`
- `VulnerabilityTest` proves the `remediationHint` field serializes under the new name without a legacy `hint` key
- All existing tests updated to reference `notices()` instead of `warnings()` and `NoticeCollector` instead of `WarningCollector`

This is a **breaking wire-format change**. Any downstream consumer parsing `warnings` from MCP tool responses must update to `notices`. The parent PR's MCP_STANDARDS v2 established `notices` as the target name and flagged the rename as pending, so consumers following the standard already expect this.

The hosted server (aiml-services) consumes contrast-mcp-core as a Maven dependency. The actual artifact-coordinate bump must wait for the breaking core release to be published. Changes were verified locally using composite build substitution.

[AIML-942]: https://contrast.atlassian.net/browse/AIML-942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/MCP_STANDARDS.md
+++ b/MCP_STANDARDS.md
@@ -211,7 +211,7 @@ Required-ness goes on the `@ToolParam(required=...)` flag, never in prose.
 Every tool response wraps in a shared envelope with exactly two message tiers. The distinction is behavioral, an agent acts differently on each.
 
 - **`errors`** mean the call failed or was invalid. Fix the call and retry. `isSuccess()` keys off this list.
-- **`notices`** are informational. Read them while interpreting the result. They carry applied defaults, failed optional enrichments, empty-result explanations, and interpretation facts. (Migration note. The wire field is named `warnings` until the AIML-942 envelope rename ships. New code should target `notices`.)
+- **`notices`** are informational. Read them while interpreting the result. They carry applied defaults, failed optional enrichments, empty-result explanations, and interpretation facts.
 
 Do not add a third tier. A distinction that does not change agent behavior is taxonomy, not signal.
 
@@ -231,7 +231,7 @@ Interpretation facts are delivered where the confusion would occur, not up front
 
 **Conditional quirks** become notices emitted only when the condition occurs on that response. An always-emitted notice is a smell, it belongs in another home.
 
-**Always-true field meaning** is carried by the field name itself. Prefer renaming a misleading field (`httpRequest` that has had sensitive headers stripped becomes `httpRequestRedacted`) over documenting it. A rename costs zero recurring tokens. When a rename is not feasible, the fallback is one clause in the description body, tagged transitional if a rename is planned.
+**Always-true field meaning** is carried by the field name itself. Prefer renaming a misleading field (`hint` containing AI-generated remediation guidance becomes `remediationHint`) over documenting it. A rename costs zero recurring tokens. When a rename is not feasible, the fallback is one clause in the description body, tagged transitional if a rename is planned.
 
 ### Output schema, end state
 
@@ -286,23 +286,23 @@ com.contrast.labs.ai.mcp.contrast.tool/
 **`PaginatedTool<P extends ToolParams, R>`** - For paginated search/list tools:
 - Template method `executePipeline()` handles pagination, validation, exceptions
 - Subclasses implement `doExecute()` returning `ExecutionResult<R>`
-- Returns `PaginatedToolResponse<R>` with items, pagination metadata, errors, warnings
+- Returns `PaginatedToolResponse<R>` with items, pagination metadata, errors, notices
 
 **`SingleTool<P extends ToolParams, R>`** - For single-item get tools:
 - Template method `executePipeline()` handles validation, exceptions
 - Subclasses implement `doExecute()` returning item or null
-- Returns `SingleToolResponse<R>` with item, errors, warnings
+- Returns `SingleToolResponse<R>` with item, errors, notices
 
 **`CursorPaginatedTool<P extends ToolParams, R>`** - For cursor/keyset-backed list tools:
 - Template method `executePipeline()` handles cursor pagination, validation, exceptions
 - Subclasses treat cursor values as opaque continuation tokens
-- Returns `CursorToolResponse<R>` with items, `nextCursor`, `hasMore`, errors, and warnings
+- Returns `CursorToolResponse<R>` with items, `nextCursor`, `hasMore`, errors, and notices
 
 ### Parameter Classes (Params Pattern)
 
 Each tool has an associated `*Params` class extending `BaseToolParams` and validating input through a `ToolValidationContext` used by composition:
 - Validates and parses input parameters
-- Collects errors and warnings via fluent API
+- Collects errors and notices via fluent API
 - Converts to SDK filter objects (e.g., `toTraceFilterForm()`)
 
 Example:
@@ -318,7 +318,7 @@ public class VulnerabilityFilterParams extends BaseToolParams {
     var ctx = new ToolValidationContext(); // composition, not inheritance
     params.severities = ctx.enumSetParam(severities, RuleSeverity.class, "severities").get();
     // ... more fluent validation on ctx
-    params.setValidationResult(ctx); // transfer errors/warnings
+    params.setValidationResult(ctx); // transfer errors/notices
     return params;
   }
 }
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/CursorPaginatedTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/CursorPaginatedTool.java
@@ -65,9 +65,9 @@ public abstract class CursorPaginatedTool<P extends ToolParams, R> extends BaseT
 
     var pagination = CursorPaginationParams.of(cursor, pageSize, getMaxPageSize());
     var params = paramsSupplier.get();
-    var collector = WarningCollector.forContext(Map.of(LoggingKeys.REQUEST_ID, requestId));
-    pagination.warnings().forEach(collector::warn);
-    params.warnings().forEach(collector::warn);
+    var collector = NoticeCollector.forContext(Map.of(LoggingKeys.REQUEST_ID, requestId));
+    pagination.notices().forEach(collector::notice);
+    params.notices().forEach(collector::notice);
 
     if (!params.isValid()) {
       logValidationError(requestId, params.errors(), pagination.cursorPresence());
@@ -105,12 +105,12 @@ public abstract class CursorPaginatedTool<P extends ToolParams, R> extends BaseT
    *
    * @param pagination validated cursor pagination params
    * @param params validated tool-specific params
-   * @param collector warning accumulator
+   * @param collector notice accumulator
    * @return cursor execution result
    * @throws Exception from downstream clients or processing
    */
   protected abstract CursorExecutionResult<R> doExecute(
-      CursorPaginationParams pagination, P params, WarningCollector collector) throws Exception;
+      CursorPaginationParams pagination, P params, NoticeCollector collector) throws Exception;
 
   private CursorToolResponse<R> handleException(
       Exception e, CursorPaginationParams pagination, String requestId, String userMessage) {
@@ -127,7 +127,7 @@ public abstract class CursorPaginatedTool<P extends ToolParams, R> extends BaseT
       HttpResponseException e,
       CursorPaginationParams pagination,
       String requestId,
-      WarningCollector collector) {
+      NoticeCollector collector) {
 
     String errorMessage = mapHttpErrorCode(e.getCode());
 
@@ -151,12 +151,12 @@ public abstract class CursorPaginatedTool<P extends ToolParams, R> extends BaseT
   private CursorToolResponse<R> buildSuccessResponse(
       CursorExecutionResult<R> result,
       CursorPaginationParams pagination,
-      WarningCollector collector,
+      NoticeCollector collector,
       long duration,
       String requestId) {
 
-    if (result.items().isEmpty() && !result.hasMore() && !collector.hasEmptyResultsWarning()) {
-      collector.warn("No results found matching the specified criteria.");
+    if (result.items().isEmpty() && !result.hasMore() && !collector.hasEmptyResultsNotice()) {
+      collector.notice("No results found matching the specified criteria.");
     }
 
     logSuccess(requestId, duration, result.items().size(), pagination.cursorPresence());
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/PaginatedToolResponse.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/PaginatedToolResponse.java
@@ -19,10 +19,10 @@ import java.util.List;
 
 /**
  * Generic paginated response wrapper for all list-returning MCP tools. Provides consistent
- * pagination metadata, error/warning separation, and timing across all endpoints.
+ * pagination metadata, error/notice separation, and timing across all endpoints.
  *
  * <p>AI agents can use {@link #isSuccess()} to determine if the request succeeded, and examine
- * {@link #errors()} for actionable problems vs {@link #warnings()} for informational messages.
+ * {@link #errors()} for actionable problems vs {@link #notices()} for informational messages.
  *
  * @param <T> The type of items in the paginated response
  * @param items The data for the current page (never null, empty list if no results)
@@ -31,7 +31,8 @@ import java.util.List;
  * @param totalItems Total count across all pages (null if unavailable or expensive to compute)
  * @param hasMorePages true if additional pages exist beyond this page
  * @param errors Validation or execution errors (empty if success) - actionable problems
- * @param warnings Non-fatal warnings (e.g., applied defaults, empty results) - informational
+ * @param notices Informational notices: applied defaults, failed optional enrichments, empty-result
+ *     explanations, interpretation facts
  * @param durationMs Execution duration in milliseconds (null for validation errors)
  */
 public record PaginatedToolResponse<T>(
@@ -41,14 +42,14 @@ public record PaginatedToolResponse<T>(
     Integer totalItems,
     boolean hasMorePages,
     List<String> errors,
-    List<String> warnings,
+    List<String> notices,
     Long durationMs) {
 
   /** Compact constructor ensures non-null, immutable lists. */
   public PaginatedToolResponse {
     items = items != null ? List.copyOf(items) : List.of();
     errors = errors != null ? List.copyOf(errors) : List.of();
-    warnings = warnings != null ? List.copyOf(warnings) : List.of();
+    notices = notices != null ? List.copyOf(notices) : List.of();
   }
 
   /**
@@ -61,14 +62,14 @@ public record PaginatedToolResponse<T>(
   }
 
   /**
-   * Creates a successful response with items and optional warnings.
+   * Creates a successful response with items and optional notices.
    *
    * @param items the response items
    * @param page page number (1-based)
    * @param pageSize items per page
    * @param totalItems total count (null if unavailable)
    * @param hasMorePages true if more pages exist
-   * @param warnings non-fatal warnings
+   * @param notices informational notices
    * @param durationMs execution time in milliseconds
    * @param <T> item type
    * @return successful paginated response
@@ -79,10 +80,10 @@ public record PaginatedToolResponse<T>(
       int pageSize,
       Integer totalItems,
       boolean hasMorePages,
-      List<String> warnings,
+      List<String> notices,
       Long durationMs) {
     return new PaginatedToolResponse<>(
-        items, page, pageSize, totalItems, hasMorePages, List.of(), warnings, durationMs);
+        items, page, pageSize, totalItems, hasMorePages, List.of(), notices, durationMs);
   }
 
   /**
@@ -101,17 +102,17 @@ public record PaginatedToolResponse<T>(
   }
 
   /**
-   * Creates an empty paginated response with a warning message.
+   * Creates an empty paginated response with a notice message.
    *
    * @param page page number (1-based)
    * @param pageSize items per page
-   * @param warningMessage informational message about empty results
+   * @param noticeMessage informational message about empty results
    * @param <T> item type
-   * @return empty response with warning
+   * @return empty response with notice
    */
-  public static <T> PaginatedToolResponse<T> empty(int page, int pageSize, String warningMessage) {
+  public static <T> PaginatedToolResponse<T> empty(int page, int pageSize, String noticeMessage) {
     return new PaginatedToolResponse<>(
-        List.of(), page, pageSize, 0, false, List.of(), List.of(warningMessage), null);
+        List.of(), page, pageSize, 0, false, List.of(), List.of(noticeMessage), null);
   }
 
   /**
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/validation/ToolValidationContext.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/validation/ToolValidationContext.java
@@ -24,8 +24,8 @@ import java.util.UUID;
 import org.springframework.util.StringUtils;
 
 /**
- * Fluent validation context for MCP tool parameters. Collects errors and warnings during
- * validation, providing AI-friendly feedback messages.
+ * Fluent validation context for MCP tool parameters. Collects errors and notices during validation,
+ * providing AI-friendly feedback messages.
  *
  * <p>Usage example:
  *
@@ -48,7 +48,7 @@ import org.springframework.util.StringUtils;
  */
 public class ToolValidationContext implements ToolParams {
 
-  private final List<String> warnings = new ArrayList<>();
+  private final List<String> notices = new ArrayList<>();
   private final List<String> errors = new ArrayList<>();
 
   /**
@@ -260,14 +260,14 @@ public class ToolValidationContext implements ToolParams {
   }
 
   /**
-   * Adds a warning if the condition is true.
+   * Adds a notice if the condition is true.
    *
    * @param condition the condition to check
-   * @param message the warning message if condition is true
+   * @param message the notice message if condition is true
    */
-  public void warnIf(boolean condition, String message) {
+  public void noticeIf(boolean condition, String message) {
     if (condition) {
-      warnings.add(message);
+      notices.add(message);
     }
   }
 
@@ -293,12 +293,12 @@ public class ToolValidationContext implements ToolParams {
   }
 
   /**
-   * Adds a warning message. Protected for use by Spec classes and subclasses.
+   * Adds a notice message. Protected for use by Spec classes and subclasses.
    *
-   * @param message the warning message
+   * @param message the notice message
    */
-  protected void addWarning(String message) {
-    warnings.add(message);
+  protected void addNotice(String message) {
+    notices.add(message);
   }
 
   @Override
@@ -312,7 +312,7 @@ public class ToolValidationContext implements ToolParams {
   }
 
   @Override
-  public List<String> warnings() {
-    return List.copyOf(warnings);
+  public List<String> notices() {
+    return List.copyOf(notices);
   }
 }
--- a/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/result/VulnerabilityTest.java
+++ b/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/result/VulnerabilityTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2025 Contrast Security
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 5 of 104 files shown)


contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/PaginatedToolResponse.java
```diff
@@ -41,14 +42,14 @@ public record PaginatedToolResponse<T>(
     Integer totalItems,
     boolean hasMorePages,
     List<String> errors,
-    List<String> warnings,
+    List<String> notices,
```
ChrisEdwards, line 45: **[HIGH]** Breaking API/wire format rename with no CHANGELOG entry

This PR renames the `warnings` record component to `notices` across all three response envelopes (`PaginatedToolResponse`, `SingleToolResponse`, `CursorToolResponse`), the `WarningCollector` class to `NoticeCollector`, `ToolParams.warnings()` to `notices()`, and `Vulnerability.hint` to `remediationHint`. Since `contrast-mcp-core` publishes to a public Maven repository, these are breaking changes for both binary compatibility (compiled consumers) and wire format (JSON consumers reading the `warnings`/`hint` keys).

The project’s prior release (2.1.0) documented a smaller breaking change (`Cve.getScore()` type change) under a "Breaking Changes" heading, and git history shows three prior instances where missing CHANGELOG entries were added as PR-review follow-ups. The `[Unreleased]` section is currently empty.

Also affects: `SingleToolResponse.java`, `CursorToolResponse.java`, `NoticeCollector.java`, `ToolParams.java`, `Vulnerability.java`

**Suggested fix:** Add a "Breaking Changes" entry to `CHANGELOG.md` under `[Unreleased]` documenting the renamed types, methods, and JSON fields.

---
*Found by: Claude (correctness-adversarial-reviewer, project-standards-reviewer, api-contract-reviewer, docs-staleness-reviewer)*

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/validation/ToolValidationContextTest.java
```diff
@@ -315,21 +315,21 @@ void requireAtLeastOne_should_pass_when_multiple_present() {
     assertThat(ctx.isValid()).isTrue();
   }
 
-  // -- warnIf tests --
+  // -- noticeIf tests --
 
   @Test
   void warnIf_should_add_warning_when_condition_true() {
```
ChrisEdwards: **[MEDIUM]** Test method names still reference removed `warnIf` method

`warnIf_should_add_warning_when_condition_true()` and `warnIf_should_not_add_warning_when_condition_false()` now call `ctx.noticeIf(...)` in their bodies but still name the old method. Every other test in this file was correctly renamed (e.g. `should_collect_multiple_notices`).

**Suggested fix:** Rename to `noticeIf_should_add_notice_when_condition_true` / `noticeIf_should_not_add_notice_when_condition_false`.

---
*Found by: Claude (project-standards-reviewer, maintainability-reviewer)*

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/CursorPaginatedTool.java
```diff
@@ -65,9 +65,9 @@ protected final CursorToolResponse<R> executePipeline(
 
     var pagination = CursorPaginationParams.of(cursor, pageSize, getMaxPageSize());
     var params = paramsSupplier.get();
-    var collector = WarningCollector.forContext(Map.of(LoggingKeys.REQUEST_ID, requestId));
-    pagination.warnings().forEach(collector::warn);
-    params.warnings().forEach(collector::warn);
+    var collector = NoticeCollector.forContext(Map.of(LoggingKeys.REQUEST_ID, requestId));
+    pagination.notices().forEach(collector::notice);
+    params.notices().forEach(collector::notice);
```
ChrisEdwards, line 70: **[MEDIUM]** Notices collected before validation/exception are lost on error paths

Notices gathered during param construction (lines 69-70) are discarded when the tool exits through `validationError()` (line 74), `handleException()`, or `handleHttpResponseException()`. These error factory methods do not accept the `NoticeCollector`, so any soft-validation notices (e.g. applied defaults, deprecated filter usage) are silently dropped from the error response.

Also affects: `PaginatedTool.java:99` (same pattern), `SingleTool.java:78`

**Suggested fix:** Pass `collector.snapshot()` through the error/exception response constructors so notices survive alongside errors.

---
*Found by: Gemini*

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/ResponseEnvelopeSerializationTest.java
```diff
@@ -0,0 +1,58 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class ResponseEnvelopeSerializationTest {
+
+  private final ObjectMapper objectMapper = new ObjectMapper();
+
+  @Test
+  void paginatedResponse_should_serialize_notices_without_legacy_warnings_field() throws Exception {
```
ChrisEdwards, line 29: **[MEDIUM]** Wire-shape regression test never exercises empty-notices case

All three test cases construct responses with non-empty notices lists. The most common runtime shape (a clean success with zero notices) is unverified. A future change adding `@JsonInclude(NON_EMPTY)` would silently omit the `notices` key from the JSON, breaking consumers that check for its presence, and this test would not catch it.

**Suggested fix:** Add a test case constructing each response type with `List.of()` for notices and assert `"notices":[]` is present in the JSON.

---
*Found by: Claude (testing-reviewer)*

MCP_STANDARDS.md
```diff
@@ -231,7 +231,7 @@ Interpretation facts are delivered where the confusion would occur, not up front
 
 **Conditional quirks** become notices emitted only when the condition occurs on that response. An always-emitted notice is a smell, it belongs in another home.
 
-**Always-true field meaning** is carried by the field name itself. Prefer renaming a misleading field (`httpRequest` that has had sensitive headers stripped becomes `httpRequestRedacted`) over documenting it. A rename costs zero recurring tokens. When a rename is not feasible, the fallback is one clause in the description body, tagged transitional if a rename is planned.
+**Always-true field meaning** is carried by the field name itself. Prefer renaming a misleading field (`hint` containing AI-generated remediation guidance becomes `remediationHint`) over documenting it. A rename costs zero recurring tokens. When a rename is not feasible, the fallback is one clause in the description body, tagged transitional if a rename is planned.
```
ChrisEdwards, line 234: **[MEDIUM]** Standards example references the rename being introduced in this same PR

The `hint` → `remediationHint` example is used as the canonical illustration of a good field rename, but it ships alongside the rename itself. Readers seeing this example won't have prior precedent to understand the principle from. The original example (`httpRequest` → `httpRequestRedacted`) was already shipped and served this purpose.

**Suggested fix:** Restore the previously-shipped `httpRequestRedacted` example or use another already-completed rename.

---
*Found by: Claude (correctness-adversarial-reviewer)*

ChrisEdwards (commented):
## Multi-model Review

This PR consistently applies the warnings→notices and hint→remediationHint renames across 100 files. The rename is mechanically correct and all tests pass. The main concern is that `contrast-mcp-core` is a published Maven library, and this PR introduces breaking API and wire-format changes with no CHANGELOG entry and no backward-compatibility mechanism, despite the project’s established convention of documenting such changes. Several test names and diagnostic strings also lag behind the rename.

<details>
<summary>Minor notes (6)</summary>

- `GetRouteCoverageToolIT.java:301`, `GetSessionMetadataToolIT.java:301`, `GetSastProjectToolIT.java:55`, etc. — Diagnostic `.as()` strings, comments, and constants (`RESOURCE_NOT_FOUND_WARNING`, `MUTUALLY_EXCLUSIVE_FILTERS_WARNING`) still say "warning" after rename to "notices". *(Claude maintainability-reviewer, testing-reviewer)*
- `docs/adr/0001-notices-tier-and-point-of-use-teaching.md` — ADR reads as still pending ("about to become") when the rename is shipping. *(Claude docs-staleness-reviewer)*
- `PaginatedToolTest`, `SingleToolTest` — Lambda parameters still named `warnings` though collector type is now `NoticeCollector`. *(Claude maintainability-reviewer)*
- `GetSastResultsTool.java:47` — `WARNING:` → `NOTICE:` in tool description may read as less urgent for a deprecation caution. *(Claude api-contract-reviewer, Gemini)*
- `AttackFilterParams:127`, `SearchAppVulnerabilitiesParams:123`, `VulnerabilityFilterParams:107` — `ctx.noticeIf(true, ...)` inside an `if` guard; the `true` literal is dead logic. *(Claude correctness-adversarial-reviewer)*
- `docs/superpowers/plans/` and `docs/superpowers/specs/` — Future tool specs still reference old `warnings`/`WarningCollector` API. *(Claude maintainability-reviewer)*
</details>

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/PaginatedToolResponse.java
ChrisEdwards: Addressed in cea83c4. Added an Unreleased Breaking Changes entry covering the public Java API and JSON wire renames for notices, NoticeCollector, ToolParams.notices(), and Vulnerability.remediationHint.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/validation/ToolValidationContextTest.java
ChrisEdwards: Addressed in cfdd538. Renamed both stale warnIf test methods to noticeIf/add_notice names and verified ToolValidationContextTest.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/CursorPaginatedTool.java
ChrisEdwards: No code change. The pre-rename contract deliberately omits informational messages on validation and general failure responses, while the standard defines notices as context for interpreting a returned result. The repository also has an integration assertion that error responses must not mix in notices. HTTP exception preservation is pre-existing behavior, not evidence that every failure path should change in this rename PR. Expanding notices onto all failures would be a semantic contract change and could distract from the actionable errors, so this remains unchanged.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/ResponseEnvelopeSerializationTest.java
ChrisEdwards: Addressed in 8d98a34. Added serialization coverage for empty notices lists on PaginatedToolResponse, SingleToolResponse, and CursorToolResponse, asserting that each emits notices:[] and never emits the legacy warnings field.

MCP_STANDARDS.md
ChrisEdwards: No code change. The httpRequest to httpRequestRedacted rename was explicitly declined because the clarity gain does not justify another public breaking change. Restoring that example would falsely document a rename that is neither planned nor shipped. The remediationHint example accurately illustrates the field-name principle and matches the only response-field rename in this PR.

ChrisEdwards:
Follow-up to @ChrisEdwards [multi-model review](https://github.com/Contrast-Security-OSS/mcp-contrast/pull/179#pullrequestreview-4815921524): addressed the tracked stale test names, constants, comments, assertion labels, ADR tense, and redundant noticeIf(true) predicates in eb2fc8b. Verified that the reported stale warnings lambda parameters are no longer present. Kept NOTICE in the deprecated SAST tool description because DEPRECATED already carries the urgency and NOTICE matches the response-tier vocabulary. The docs/superpowers planning archive is gitignored local history rather than tracked PR content, so it was not force-added or represented as changed. Full spotlessCheck/check, Checkstyle, unit tests, publication metadata, and JaCoCo verification pass.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/ResponseEnvelopeSerializationTest.java
```diff
@@ -0,0 +1,73 @@
...
+  private void assertNoticesWireShape(Object response) throws Exception {
+    var json = objectMapper.writeValueAsString(response);
+
+    assertThat(json).contains("\"notices\":[").doesNotContain("\"warnings\"");
+  }
+
+  private void assertEmptyNoticesWireShape(Object response) throws Exception {
+    var json = objectMapper.writeValueAsString(response);
+
+    assertThat(json).contains("\"notices\":[]").doesNotContain("\"warnings\"");
```
Alex-Contrast, line 71: This assertion pins Jackson's default inclusion policy rather than the wire format the PR is trying to lock down.

`"notices":[]` is purely a claim about property inclusion. The test uses a bare `new ObjectMapper()` (line 26), not the mapper Spring AI actually serializes tool responses with. I checked and there is currently no `spring.jackson.*` property and no global Jackson customization in the repo, so today the two agree. But if anyone later sets `spring.jackson.default-property-inclusion=non_empty`, production stops emitting the field entirely and this test stays green.

Worth injecting the application's `ObjectMapper` here so the test fails when the real wire shape changes. This is the one residual risk I could not rule out by reading the diff, since everything else about the rename verifies clean.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/ResponseEnvelopeSerializationTest.java
```diff
@@ -0,0 +1,73 @@
...
+        PaginatedToolResponse.success(List.of("item"), 1, 50, 1, false, List.of(), 1L));
+    assertEmptyNoticesWireShape(SingleToolResponse.success("item", List.of()));
+    assertEmptyNoticesWireShape(
+        CursorToolResponse.success(List.of("item"), 50, null, false, List.of(), 1L));
+  }
+
+  private void assertNoticesWireShape(Object response) throws Exception {
+    var json = objectMapper.writeValueAsString(response);
+
+    assertThat(json).contains("\"notices\":[").doesNotContain("\"warnings\"");
```
Alex-Contrast, line 65: Minor durability point. `contains("\"notices\":[")` depends on Jackson emitting no space after the colon, so these tests can fail for a formatting reason that has nothing to do with the field name they exist to verify.

Parsing to a `JsonNode` and asserting `has("notices")` and `!has("warnings")` gives the same coverage and is immune to spacing and field ordering.

Also, `assertNoticesWireShape(Object response)` typed as `Object` means a future refactor that passes the wrong object still compiles. A bounded generic or three explicit overloads would catch that at compile time.

Alex-Contrast (approved):

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/result/VulnerabilityTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.jupiter.api.Test;
+
+class VulnerabilityTest {
+
+  private final ObjectMapper objectMapper = new ObjectMapper();
+
+  @Test
+  void serialization_should_expose_remediationHint_without_legacy_hint() throws Exception {
+    var vulnerability =
+        new Vulnerability(
```
Alex-Contrast, line 31: Verified this is correct today. I diffed the record component order between `main` and this branch and `remediationHint` is component one in both, so the first argument does land on the field under test.

The durability concern is that this is an 18-component record built positionally with 11 nulls and empty lists. If someone later inserts another `String` component near the front, this test silently starts asserting on a different field and still passes as long as a value lands there. A brief comment naming the component, or a small named-local, is enough to make the coupling visible.

For context on the positional risk more broadly, `SingleToolResponse` has two adjacent `List<String>` components (`errors`, `notices`) and `PaginationParams` has four adjacent `int`s, so a swap in any of those compiles clean and produces wrong data. I checked component order on all five renamed records against `main` and every one is unchanged, so nothing is wrong here, but that class of mistake is invisible to the compiler.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/validation/ToolValidationContext.java
```diff
@@ -260,14 +260,14 @@ public void mutuallyExclusive(
...
   /**
-   * Adds a warning if the condition is true.
+   * Adds a notice if the condition is true.
    *
    * @param condition the condition to check
-   * @param message the warning message if condition is true
+   * @param message the notice message if condition is true
    */
-  public void warnIf(boolean condition, String message) {
+  public void noticeIf(boolean condition, String message) {
```
Alex-Contrast, line 268: Not introduced by this PR, but the rename makes it easier to trip over so flagging it while the code is fresh.

There are now two notice sinks that share a name and have different contracts. `NoticeCollector.notice()` throws on null and silently drops blank strings. This path adds unconditionally (also at line 301), and `notices()` at line 315 is a bare `List.copyOf`. Both feed the same `notices` array on the wire.

Consequences: a blank message from the validation path emits `""` as a notice element, and a null one throws from `List.copyOf` at snapshot time rather than at the call site where the bug is.

Before the rename these looked like two unrelated helpers. Now they look like the same operation, so aligning the null and blank handling would be worth a follow-up.

I confirmed the behavior of the three call-site refactors in this PR is unchanged. `noticeIf(cond, m)` replacing `if (cond) { warnIf(true, m) }` in `SearchAppVulnerabilitiesParams` and `VulnerabilityFilterParams` is identical, and the `AttackFilterParams` move out of the null branch keeps both the condition and the notice ordering the same.

Alex-Contrast (commented):
Reviewed for correctness. **No correctness defects found in the rename.** Four inline comments, all test-durability improvements that I do not think need to gate the merge.

Because this is 104 files of mostly mechanical change, I verified it mechanically rather than by eye. Recording what is already covered so nobody redoes it.

| Check | Method | Result |
|---|---|---|
| Hidden reorderings | Normalized the rename mapping over every removed line and compared to added lines, run both order-sensitive and order-insensitive | Both passes return the identical 31 non-rename hunks out of 4,770 diff lines. Nothing anywhere was reordered, only substituted |
| Positional record safety | Diffed component order `main` vs HEAD for `Vulnerability` (18 components), `PaginatedToolResponse`, `SingleToolResponse`, `CursorToolResponse`, `PaginationParams`, `CursorPaginationParams` | All unchanged. Factory bodies still pass `errors` and `List.of()` in the same slots at `PaginatedToolResponse:101,115,129` |
| Accidental literal rewrites | Extracted every string literal in `src/main` from both sides of the diff and diffed the multisets | Exactly one changed, `"warning message must not be null"` to `"notice message must not be null"` at `NoticeCollector:114`. No user-facing notice text, no API query parameter, no JSON key |
| Rename completeness | Repo-wide sweep across Java, resources, and config | Zero residual `warnings` / `warnIf` / `WarningCollector`. Surviving `warn` hits are all legitimate (SLF4J `log.atWarn`, `WARN` in the `RunBookEnum` CEF samples, the log-level set in `ServerFilterParams`) |
| Empty-results marker chain | Traced `noticeForEmptyResults` to `emptyResultsNoticeRecorded` to its readers | Intact at `PaginatedTool:194` and `CursorPaginatedTool:158`, blank-skip semantics preserved at `NoticeCollector:125-130`, package-private accessor still resolves |
| Logic refactors | Compared all three against `main` with full context | Behavior-identical. Details in the `ToolValidationContext` comment |
| `hint` scoping | Checked `VulnLight` and the `hints/` package | `VulnLight` has no `hint` component, so the search and get shapes stay consistent. Generator internals correctly untouched |
| Tests | `./gradlew test` under JDK 21 | 1034 tests, 0 failures. New tests executed (`ResponseEnvelopeSerializationTest` 4, `VulnerabilityTest` 1, `NoticeCollectorTest` 15) |

**On the wire-format proof.** I initially thought the absence of an end-to-end params-to-JSON test was worth holding the PR for, and I was wrong about that. The two halves compose. The base tool tests prove a tool populates `notices()`, and `ResponseEnvelopeSerializationTest` proves a record carrying `notices` serializes under that name, and the records *are* the wire types. There is no untested seam between those two facts. The real residual risk is narrower and specific, whether Spring AI's mapper matches the bare `ObjectMapper` the new test uses, which is the top inline comment.

Recommending approve once the mapper-fidelity point is either addressed or consciously accepted.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/ResponseEnvelopeSerializationTest.java
ChrisEdwards: Addressed on the tip stacked PR, #183, in commit 1e1eab4.

I revalidated this against the tip branch before changing it. The fidelity concern was still valid, although the application's injected ObjectMapper is not the serializer used for tool results: Spring AI's DefaultToolCallResultConverter delegates to its static JsonParser. The test now serializes through JsonParser.toJson(...) and parses with JsonParser.getObjectMapper(), so it exercises the actual production conversion path and will detect changes to that mapper's inclusion policy.

The focused test passes, and the commit hook also completed the full unit-test and Checkstyle tasks successfully.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/ResponseEnvelopeSerializationTest.java
ChrisEdwards: Addressed on the tip stacked PR, #183, in commit 1e1eab4 after confirming the concern still applied there.

The test now parses the production JsonParser output into JsonNode and asserts the semantic shape with has("notices"), !has("warnings"), an array-type check, and a zero-size check for empty notices. This removes whitespace and field-order coupling. I also replaced the Object-typed response helper with typed serialization overloads for PaginatedToolResponse, SingleToolResponse, and CursorToolResponse.

The focused test passes, and the commit hook completed the full unit-test and Checkstyle tasks successfully.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/result/VulnerabilityTest.java
ChrisEdwards: No code change. I revalidated this test on the tip stacked branch and the positional-constructor concern remains non-blocking rather than a latent false-positive.

Adding a record component changes the canonical constructor arity and breaks compilation until the call site is updated. If remediationHint moves to another position, the assertion's exact expected value no longer lands in the remediationHint property and the test fails. A named local would improve readability but would not add durability or compiler protection, so I am leaving the focused serialization test unchanged.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/validation/ToolValidationContext.java
ChrisEdwards: Deferred to follow-up AIML-1266 (https://contrast.atlassian.net/browse/AIML-1266), with local bead mcp-77hk linked as discovered-from the active AIML-942 migration work.

I revalidated the behavior on the tip stacked branch. The two APIs still have inconsistent null/blank contracts, so the consistency concern is valid. The claimed blank wire element does not occur in current tool pipelines, however: validation notices pass through collector::notice before successful envelopes are built, which filters blanks. The remaining issue is chiefly contract consistency and failure locality, because null currently fails later in ToolValidationContext.notices() via List.copyOf.

The follow-up scopes a shared policy, centralized ToolValidationContext insertion, focused null/blank/false-condition tests, and verification that the response wire behavior remains unchanged. No semantic validation change is being mixed into this rename/description stack.

PR #188, opened 2026-08-04
AIML-1285 Release stdio MCP server v2.2.0

ChrisEdwards:
**Jira:** [AIML-1285](https://contrast.atlassian.net/browse/AIML-1285)

## Summary

Adds release tooling for the v2.2.0 release: a new `/update-changelog` skill that automates pre-release changelog authoring, hardening for the `/test-mcp-server` skill, and the finished changelog itself.

- `/update-changelog` diffs the last release tag against main, drafts entries PR-by-PR, audits with a read-only subagent, and lands the result after human approval
- `/test-mcp-server` now supports `--model` for tester selection, writes per-tool result files, and salvages partial results when the orchestrator exits early
- CHANGELOG.md documents all user-facing changes since v2.1.0

## Why This Change Exists

The changelog was previously written by hand, which was error-prone and easy to forget. The v2.1.0 release shipped without a changelog until it was backfilled after the fact. The `/update-changelog` skill automates the enumeration, classification, and auditing steps so nothing gets missed.

The `/test-mcp-server` skill had three pain points: the orchestrator sometimes gets killed mid-run (context limits or timeouts) and all individual tool results were lost, the 600-second background output capture ceiling cut off long regular runs before they finished, and there was no way to use a different model for tester subagents.

## Approach

The changelog skill follows a structured flow: orient on tags and headings, enumerate PRs in the range, classify each as user-facing or internal, draft entries in the existing house style, audit with a read-only subagent, and present the draft for human approval before writing anything. It supports both current-branch mode (changes ride the release branch) and new-branch mode (worktree-based, separate PR).

During this run, the notices/remediationHint wire rename was reclassified from Breaking Changes to Improvements. MCP server consumers (AI agents) discover response shapes dynamically, so a field rename cannot break them. This dropped the suggested version from v3.0.0 to v2.2.0. A subagent measured the actual token reduction from the AIML-942 description rewrite (55% cut in tool descriptions, ~650 fewer tokens per session), and that concrete metric was folded into the changelog entry.

For test-mcp-server, per-tool result files solve the partial-run problem: each tester writes its verdict to `$RESULTS_DIR/<tool-name>.txt`, so an early exit still produces usable output. The `--model` flag passes through to the agent definition, and `CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS=0` removes the background capture timeout.

### Release tooling
- `.claude/skills/update-changelog/SKILL.md` (new) — the full skill definition with steps, ground rules, audit subagent prompt template, and house style reference
- `.claude/skills/test-mcp-server/SKILL.md` — documents `--model` flag, updates launch command with `CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS=0`, corrects run-time estimate
- `.claude/skills/test-mcp-server/run.sh` — parses `--model` flag, creates per-tool results directory, adds early-exit salvage block, allows Bash tool for result file writes

### Changelog and docs
- `CHANGELOG.md` — full `[Unreleased]` section covering bug fixes, improvements, and documentation since v2.1.0
- `CLAUDE.md` — adds `/update-changelog` skill description to the Agent Skills section
- `RELEASING.md` — adds "Pre-release Changelog Update" section and inserts the changelog step into the release checklist

## Testing

- Both skills are internal development tooling, not shipped code. The changelog skill was exercised during this session to produce the v2.2.0 changelog (its first real run).
- The test-mcp-server changes are tested by running `/test-mcp-server` as part of the pre-release process.
- No production code changed; no unit or integration tests affected.

[AIML-1285]: https://contrast.atlassian.net/browse/AIML-1285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.claude/skills/test-mcp-server/SKILL.md
+++ b/.claude/skills/test-mcp-server/SKILL.md
@@ -17,20 +17,25 @@ on demand only.
 
 ## Usage
 
-- `/test-mcp-server` runs the in-depth (regular) pass.
+- `/test-mcp-server` runs the in-depth (regular) pass with Sonnet testers.
 - `/test-mcp-server smoke` runs a fast pass that just checks each tool's main use case.
 - `/test-mcp-server regular <focus>` runs in-depth with a nudge, e.g. `/test-mcp-server regular focus on the server tools`.
+- `/test-mcp-server --model claude-opus-4-6` runs with Opus testers instead of Sonnet.
+- `--model <id>` can be combined with any mode, e.g. `/test-mcp-server smoke --model claude-opus-4-6`.
 
 ## What to do
 
 1. Work out the mode and focus from the arguments. Default mode is regular. If the
    first word is not `smoke` or `regular`, treat the whole argument as focus text on
    a regular run.
-2. Launch the helper in the background, since a full regular run can take several
-   minutes, and wait for the completion notification rather than polling:
+2. Launch the helper in the background so the user can keep chatting. A regular
+   run spawns one tester per tool and can take 15-20 minutes. Set
+   `CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS=0` in the command so the background
+   output capture waits indefinitely instead of cutting off at the default
+   600-second ceiling:
 
    ```
-   bash .claude/skills/test-mcp-server/run.sh <mode> <focus>
+   CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS=0 bash .claude/skills/test-mcp-server/run.sh <mode> <focus>
    ```
 
 3. When it finishes, read its output and show the human the report as the script
--- a/.claude/skills/test-mcp-server/SKILL.md
+++ b/.claude/skills/test-mcp-server/SKILL.md
@@ -25,9 +25,10 @@ on demand only.
 
 ## What to do
 
-1. Work out the mode and focus from the arguments. Default mode is regular. If the
-   first word is not `smoke` or `regular`, treat the whole argument as focus text on
-   a regular run.
+1. Work out the mode and focus from the arguments. If the first word is `smoke` or
+   `regular`, use it as the mode and treat the rest as focus text. If no mode was
+   provided, ask the user which mode to run (smoke or regular) before launching.
+   Always pass an explicit mode to the script.
 2. Launch the helper in the background so the user can keep chatting. A regular
    run spawns one tester per tool and can take 15-20 minutes. Set
    `CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS=0` in the command so the background
--- a/.claude/skills/test-mcp-server/run.sh
+++ b/.claude/skills/test-mcp-server/run.sh
@@ -31,19 +31,27 @@
 set -euo pipefail
 
 # --- parse mode, model, and focus -----------------------------------------
-# Default to regular. Only consume the first word if it is an explicit mode;
-# otherwise the whole argument is focus text on a regular run.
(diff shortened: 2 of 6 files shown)


ChrisEdwards:
Even though the one of the main api objects changed, I did not bump this a major version. Its an mcp server, so the AI adapts and this does not really break it. If you consume this with a java app directly, then yes, it would be breaking, but no one should be doing that. That is not its intended use. Additionally, we just did a major bump for the split to a monorepo. Then we will be upgrading to spring ai soon. I just can't see changing the major version 3 releases in a row.

JacobMagesHaskinsContrast (approved):

Alex-Contrast:
# Review Findings — AIML-1285

## Findings

### Medium (should fix — correct on the normal path, wrong on a plausible one)

* [.claude/skills/test-mcp-server/run.sh:148] **Problem:** When `ORCH_EXIT` is non-zero and no `.txt` files were created in `$RESULTS_DIR`, `ls "$RESULTS_DIR"/*.txt` fails with exit code 2 because the glob pattern fails to match any files. Under `set -euo pipefail`, command substitution inside a pipeline evaluates this non-zero exit code and aborts script execution immediately.
**Impact:** When the orchestrator exits early or fails before saving any tool results (e.g., startup failure, invalid flags, missing tools, or early process termination), `run.sh` crashes with a pipeline exit error instead of executing the fallback log statement `"Orchestrator exited with code $ORCH_EXIT and no individual results were saved."`
**Fix:** Replace `ls "$RESULTS_DIR"/*.txt 2>/dev/null` with `find "$RESULTS_DIR" -maxdepth 1 -name "*.txt"`, which exits cleanly with status 0 even when no files match:
`RESULT_COUNT="$(find "$RESULTS_DIR" -maxdepth 1 -name "*.txt" | wc -l | tr -d ' ')"`

### Low (worth fixing — minor correctness, robustness, or clarity risk)

* [.claude/skills/test-mcp-server/run.sh:37] **Problem:** If `--model` is supplied at the end of the argument list without a value, `$2` is unbound and `shift 2` fails when `$#` is less than 2.
**Impact:** Under `set -euo pipefail`, passing `--model` without specifying a model name triggers an unhandled `unbound variable` error or invalid `shift` error rather than a descriptive command usage message.
**Fix:** Validate `$# -ge 2` before attempting to assign `$2` and shift arguments, or output an error message and exit.

### Nit (non-blocking — style, naming, redundancy)

* [.claude/skills/test-mcp-server/run.sh:27] **Problem:** Mode selection only inspects `$1` prior to option processing. If options such as `--model <id>` precede the mode keyword (e.g., `/test-mcp-server --model claude-opus-4-6 smoke`), `MODE` defaults to `"regular"` and `"smoke"` is appended to `FOCUS`.
**Fix:** Move positional mode keyword detection into the argument parsing loop or process options before evaluating positional mode arguments.

ChrisEdwards:
Thanks Alex — all three addressed in 56652cd:

1. **Medium (ls glob failure)**: Replaced `ls *.txt` with `find -maxdepth 1 -name '*.txt'` so it exits 0 on no matches under `pipefail`.
2. **Low (--model missing value)**: Added a `$# -lt 2` guard that prints a usage message and exits.
3. **Nit (mode ordering)**: Made mode a required first argument in the script. The skill layer now asks the user which mode to run when none is provided, so the script always gets an explicit mode. Added a comment noting mode must come first.

PR #186, opened 2026-08-04
[STACKED] AIML-942 Migrate remaining tool descriptions

ChrisEdwards:
> ⚠️ **DO NOT MERGE** - This PR is stacked on #184. Merge that first.
>
> **Stack Context**
> - Parent PR: #184
> - Parent branch: `AIML-942-add-server-instructions`
> - Promotion target: `AIML-942-migrate-vulnerability-tool-descriptions`

**Jira:** [AIML-942](https://contrast.atlassian.net/browse/AIML-942)

## Summary

Migrate all remaining (non-vulnerability) tool descriptions to the token-efficient format defined by MCP_STANDARDS.md, and move response-shape warnings from static description prose to runtime notices that appear only when the data warrants them.

- Tool descriptions across 14 source files trimmed to necessity-gated richness, cutting ~300 lines of prose
- Response-shape caveats (virtual patches, CVSS v2 scores, null Protect/agent state, session-filtered environments) converted to conditional runtime notices
- Null-safety fix for SDK returning null HTTP request response in `GetVulnerabilityTool`
- Stale `trauma_guard` hook removed from project settings

## Why This Change Exists

The first PR in this stack (#183) migrated vulnerability tool descriptions and the second (#184) added server-level instructions as a home for conventions shared across all tools. This PR completes the migration for every remaining tool: applications, attacks, coverage, libraries, SAST, and servers.

Several tools carried response-shape warnings in their static descriptions that only applied to specific data shapes (e.g., virtual patch entries, CVEs with only CVSS v2 data, servers where Protect state is unknown). These warnings consumed tokens on every call even when irrelevant. Moving them to conditional notices means agents see the warning only when the response actually contains the edge case.

The "requires admin permissions" claim on `search_vulnerabilities` was investigated during this work and found to be false. The endpoint requires only org membership (any role from viewer up), and EAC scoping returns empty results for unauthorized apps rather than 403. That sentence was deleted, and the uniform EAC-visibility fact moved to server instructions in PR #184.

## Approach

Each tool was migrated following the same pattern established in PRs #183 and #184:

1. Trim the `@Tool` description to its necessity-gated budget (get tools ~40 words, search tools ~120 words)
2. Move response-field inventories out (agents learn these from the response itself)
3. Remove "Related tools" footers (cross-references now live inline as discovery pointers or in server instructions)
4. Push parameter-specific semantics (filter logic, valid values, mutual exclusivity) down to `@ToolParam` descriptions
[...]

PR #189, opened 2026-08-05
AIML-942 Enforce per-tool description word budgets

ChrisEdwards:
**Jira:** [AIML-942](https://contrast.atlassian.net/browse/AIML-942)

## Summary
Adds an automated build gate that fails when any `@Tool` description exceeds its per-verb word ceiling, replacing manual reviewer counting with a parameterized unit test.

- All 13 tool descriptions verified under ceiling, allowlist starts empty
- New tools are discovered automatically via classpath scan, no test maintenance needed
- MCP_STANDARDS.md updated to reference the test and remove stale future-tense language

## Why This Change Exists
MCP_STANDARDS.md defines word ceilings per verb shape (search/list 150, get 40, update 60) to keep tool descriptions cheap in agent context windows. Until now, the only enforcement was reviewers counting words by hand, which was error-prone and easy to skip. After PR #186 migrated all 13 core descriptions to fit within budgets, this PR locks the door so they stay within budgets going forward.

## Approach
A `@ParameterizedTest` discovers every `@Tool`-annotated method under the tool package via Spring's `ClassPathScanningCandidateComponentProvider`, resolves each tool's verb prefix to a ceiling, and asserts the whitespace-delimited word count of the description body stays at or below that ceiling. This means newly added tools are picked up with zero test edits.

The reflection helper (`ToolTestReflection`) was copied from the aiml-services sibling rather than published as a shared test artifact, because a shared artifact would couple every gate adjustment to a core release cycle. Each copy carries a cross-reference comment.

An empty allowlist with tripwires handles edge cases: stale entries (tool now under ceiling) fail the test, entries naming unknown tools fail, and allowlist justifications must reference a specific sibling tool (sibling-disambiguation is the only valid overage reason). An unknown verb prefix also fails, forcing a ceiling decision in MCP_STANDARDS.md before any new verb ships unbudgeted.

### Test infrastructure
- `ToolTestReflection.java` (new) — classpath-scanning helper that discovers `@Tool` methods by package, ported from aiml-services
- `ToolDescriptionBudgetTest.java` (new) — parameterized test with per-verb ceilings, allowlist with stale-entry and unknown-tool tripwires, and a sanity floor asserting at least 13 tools are found

### Documentation
- `MCP_STANDARDS.md` — Enforcement section now names `ToolDescriptionBudgetTest` as the enforcement mechanism, removes the "until those tests land" future-tense sentence, and removes the catalog-total cap claim (that cap was dropped as friction-without-signal)

## Testing
- `ToolDescriptionBudgetTest` runs as part of the standard `make check-test` suite
- Red-checked by temporarily pushing `get_vulnerability` over its 40-word ceiling and confirming the failure message carries tool name, count, and ceiling
- Red-checked the stale-allowlist tripwire by adding a fake entry for `get_protect_rules` (18 words, well under the 40-word get_ ceiling) and confirming it fails
- Full repo checks pass (1,066 tests, coverage floors, static analysis)

[AIML-942]: https://contrast.atlassian.net/browse/AIML-942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/ToolDescriptionBudgetTest.java
+++ b/contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/ToolDescriptionBudgetTest.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2026 Contrast Security
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.contrast.labs.ai.mcp.contrast.tool;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.springframework.ai.tool.annotation.Tool;
+
+/** Enforces the per-tool description budgets defined in {@code MCP_STANDARDS.md}. */
+class ToolDescriptionBudgetTest {
+
+  private static final String TOOL_PACKAGE = "com.contrast.labs.ai.mcp.contrast.tool";
+  private static final int SEARCH_WORD_CEILING = 150;
+  private static final int LIST_WORD_CEILING = 150;
+  private static final int GET_WORD_CEILING = 40;
+  private static final int UPDATE_WORD_CEILING = 60;
+
+  // There were 13 core tools on 2026-08-04. Additions should not change this sanity floor.
+  private static final int MINIMUM_TOOL_COUNT = 13;
+
+  private static final Map<String, AllowlistEntry> ALLOWLIST = Map.of();
+  private static final List<Method> TOOL_METHODS = ToolTestReflection.toolMethods(TOOL_PACKAGE);
+
+  static Stream<Arguments> toolMethods() {
+    return TOOL_METHODS.stream()
+        .map(method -> Arguments.of(method.getAnnotation(Tool.class).name(), method));
+  }
+
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("toolMethods")
+  void description_should_not_exceed_its_verb_budget(String toolName, Method toolMethod) {
+    var budget = budgetFor(toolName);
+    var wordCount = wordCount(toolMethod.getAnnotation(Tool.class).description());
+    var allowlistEntry = ALLOWLIST.get(toolName);
+
+    if (allowlistEntry != null) {
+      assertThat(wordCount)
+          .as("%s is now within its base ceiling; remove the stale allowlist entry", toolName)
+          .isGreaterThan(budget.ceiling());
+      assertThat(justificationNamesSibling(toolName, allowlistEntry.justification()))
+          .as(
+              "%s allowlist justification must name the sibling tool it disambiguates against",
+              toolName)
+          .isTrue();
+    }
+
+    var effectiveCeiling =
+        allowlistEntry == null ? budget.ceiling() : allowlistEntry.grantedCeiling();
+    var ceilingSource =
+        allowlistEntry == null
+            ? budget.prefix() + " prefix"
+            : "reviewed allowlist; base " + budget.prefix() + " prefix ceiling " + budget.ceiling();
+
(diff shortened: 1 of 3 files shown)


contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/ToolDescriptionBudgetTest.java
```diff
@@ -0,0 +1,171 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.springframework.ai.tool.annotation.Tool;
+
+/** Enforces the per-tool description budgets defined in {@code MCP_STANDARDS.md}. */
+class ToolDescriptionBudgetTest {
+
+  private static final String TOOL_PACKAGE = "com.contrast.labs.ai.mcp.contrast.tool";
```
Alex-Contrast, line 36: **High** — the gate does not cover the whole agent-visible tool catalog.

**Problem:** This scans `com.contrast.labs.ai.mcp.contrast.tool` from a test living in `contrast-mcp-core`. There are 14 `@Tool` methods in this repository, not 13. The fourteenth is `get_scan_results` (`contrast-mcp-stdio-app/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/sast/GetSastResultsTool.java:41`). `contrast-mcp-stdio-app` declares `implementation project(':contrast-mcp-core')`, so the dependency runs stdio-app to core and not the reverse, and core's test classpath cannot contain that class. `ClassPathScanningCandidateComponentProvider` therefore cannot discover it. It sits in the same package namespace (`...contrast.tool.sast`), so nothing about the package name signals it is out of reach.

**Impact:** `get_scan_results` is 102 words against the 40-word `get_` ceiling, and it ships to agents ungated. Its description carries precisely the three things AIML-942 set out to remove, a standalone `Related tools:` footer, a full response-field inventory, and usage examples on a simple tool. Three consequences follow:

1. "All 13 tool descriptions verified under ceiling" describes core, not the agent-visible catalog.
2. `MCP_STANDARDS.md` now names `ToolDescriptionBudgetTest` as the enforcement mechanism, which overstates its reach.
3. Automatic discovery holds only within core, so a future tool added to `contrast-mcp-stdio-app` escapes the budget silently.

`MINIMUM_TOOL_COUNT = 13` with `hasSizeGreaterThanOrEqualTo` cannot catch any of this, because a tool that is invisible to the scan does not lower the count. The floor guards total scan failure and never partial reach.

**Fix:** Two parts, and the second matters more than the first. Bring `get_scan_results` under budget, or record it in `ALLOWLIST` with a justification. Then close the reach gap so this cannot recur, either by moving the gate to a module that sees every tool, or by asserting the discovered count against an independently derived total so that a tool outside the scanned classpath fails the build instead of passing invisibly.

Two notes on how this was established. The basis is the module layout and the dependency direction in `contrast-mcp-stdio-app/build.gradle` plus the word count of the annotation body, not an observed build failure, so if you believe the scan does reach stdio-app on CI's classpath please say so and I will recheck. Separately, the hosted server's boundary gate in `aiml-services` asserts exactly one core tool class absent from its runtime classpath, `tool/sast/GetSastResultsTool`. The same class being the odd one out in both repos looks like the same underlying seam rather than a coincidence.

Alex-Contrast (commented):
Reviewed the budget gate. One finding, inline, and it goes to the gate's coverage rather than its logic.

The short version is that the gate scans `contrast-mcp-core` only, while a fourteenth `@Tool` method lives in `contrast-mcp-stdio-app` and is roughly 2.5x over its ceiling. The sanity floor cannot detect that, because an undiscovered tool does not reduce the count.

Checked and found sound, recorded here so it does not get re-litigated:

- The ported `ToolTestReflection` helper. Scans, filters on `@Tool`, sorts by tool name, deterministic. No defect in the port, and the copy-rather-than-share decision is stated with a cross-reference comment in each copy, so I am not going to argue for extraction.
- Unknown verb prefixes. `budgetFor` throws `AssertionError` and is pinned by its own test. No escape hatch.
- The stale-entry and unknown-tool tripwires. Both correctly shaped.
- `wordCount` on a blank description. Correctly returns 0 here. Worth knowing the `aiml-services#328` sibling returns 1 for the same input, so the two hand-synced copies have diverged on that point and this one is the better of the two.

One thing deliberately not raised as a finding: `justificationNamesSibling` has no null guard on `justification`, so a hand-written `new AllowlistEntry(n, null)` would throw NPE instead of failing with the worded assertion. `ALLOWLIST` is empty and the call is reached only when an entry exists, and the test fails either way, so this is only about error-message quality on an unreachable branch. Fold it in if you are touching the file anyway.

Produced with model assistance and then verified by hand against the repo. Posting findings, not a verdict, so approval is not being withheld or granted here.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/ToolDescriptionBudgetTest.java
ChrisEdwards: Declining — `get_scan_results` is scheduled for deprecation, so we won't bring it under budget or extend the gate to cover it. The reach gap is real but will close when the tool is removed. Appreciate the thorough analysis; the observation about the classpath boundary and the sanity floor's blind spot is well taken and will inform the deprecation cleanup.

dougj-contrast (approved):

PR #207, opened 2026-08-07
AIML-1329 Override mapHttpErrorCode for 500 in ListApplicationsByCveTool

ChrisEdwards:
## Summary

- Overrides `mapHttpErrorCode` in `ListApplicationsByCveTool` to return a CVE Shield-aware message for HTTP 500 instead of the generic "Narrow filters or reduce page size" guidance, which makes no sense for a tool with neither filters nor pagination.
- The upstream endpoint returns 500 (not 404) for CVEs unknown to the SCA library data. The new message explains this and directs the user to `search_cves` or `get_cve_impact` as alternatives.
- All other status codes (401, 403, 404, 429, 502, 503, default) delegate to `super.mapHttpErrorCode()` unchanged.

## Test plan

- [x] New test: 500 returns the CVE Shield guidance message with no response-body or path leak
- [x] New test: 502 still returns the base generic message (proves override does not widen)
- [x] `make check-test` passes (1,078 tests, checkstyle, spotless, coverage floors)
- [ ] Upstream fix tracked in TS-43426; this override remains correct after that fix since 404 already maps cleanly

## Jira

[AIML-1329](https://contrast.atlassian.net/browse/AIML-1329)

[AIML-1329]: https://contrast.atlassian.net/browse/AIML-1329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
@@ -50,6 +50,12 @@ import org.springframework.stereotype.Service;
 @Slf4j
 public class ListApplicationsByCveTool extends SingleTool<ListApplicationsByCveParams, CveData> {
 
+  private static final int HTTP_INTERNAL_SERVER_ERROR = 500;
+  private static final String INTERNAL_SERVER_ERROR_MESSAGE =
+      "The service returned an error. This happens for CVEs the SCA library data does not"
+          + " recognize, including CVEs that exist only in NorthStar CVE Shield data. Verify the"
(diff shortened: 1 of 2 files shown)


Alex-Contrast (approved):

Checked the thing this change lives or dies on, that a 500 is not quietly turned into a "not found". It isn't, the override returns an error result with isSuccess() false and a populated errors(), so no caller can read it as an empty application list. The 502 test genuinely proves the override does not widen, since it asserts the unchanged base message rather than just asserting some message.

Also confirmed the premise the change rests on, that the only HttpResponseException reaching mapHttpErrorCode comes from getApplicationsByCve because enrichment failures are absorbed by collector.tryRun. If that were not true this override would misattribute unrelated 500s. It holds.

Nice touch that the override stays correct after the upstream TS-43426 fix lands rather than needing to be reverted.

PR #208, opened 2026-08-07
AIML-1331 Document lastSeen=0 sentinel and status staleness in list_applications_by_cve

ChrisEdwards:
## Summary

AI agents misread two things in the `list_applications_by_cve` response:

1. **lastSeen of 0** - Agents treat this as January 1970 or missing data. It actually means the application has never been observed running (typically a static or SCA-only upload). TeamServer sends 0 for these, and the field is a primitive long, so 0 always serializes.

2. **Server status staleness** - An agent reported "all 20 servers are OFFLINE" from this tool while `search_servers` showed one of them online. The status in this response is TeamServer's last-known state and can lag.

## What changed

**list_applications_by_cve**
- The tool now emits a notice naming the affected applications when any app in the response has `lastSeen` of 0. The notice only appears when the condition is present, following the MCP_STANDARDS.md rule that conditional quirks belong in notices, not the description body.
- The tool description now notes that `lastSeen` and server status reflect last-known agent reports and can lag live state, pointing agents to `search_servers` for fresher data.

**search_applications**
- The same lastSeen=0 sentinel exists here too, rendered even worse as a 1970 epoch date in `lastSeenAt`. The tool now emits the same conditional notice for affected applications.

Both tools already used `collector.notice()` for other conditional facts (CVSS v2 scores, empty results), so this follows the established pattern.

## Test plan

- [x] Unit tests for the new notices in both tools (positive and negative cases)
- [x] `make check-test` passes (1082 tests, 0 failures, all coverage floors met)
- [x] `make verify` passes for both changed tools (pre-existing failures in two unrelated tools are data-dependent against the live org)

Stacked on #207 (AIML-1329). Ships together in one core release plus a pin bump in aiml-services.

### DIFF
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/application/SearchApplicationsTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/application/SearchApplicationsTool.java
@@ -119,9 +119,28 @@ public class SearchApplicationsTool
       return ExecutionResult.empty();
     }
 
+    noticeNeverObservedApps(response.getApplications(), collector);
+
     var applications = response.getApplications().stream().map(this::toApplicationData).toList();
 
     return ExecutionResult.of(applications, response.getCount());
+  }
+
+  // TeamServer sends last_seen 0 for applications that have never reported agent activity, which
+  // renders as the 1970 Unix epoch in lastSeenAt (Jira AIML-1331).
+  private static void noticeNeverObservedApps(
+      List<Application> applications, NoticeCollector collector) {
+    var neverObserved =
+        applications.stream()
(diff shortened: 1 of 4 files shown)


Alex-Contrast (approved):

The notice fires only when an affected application is present and names them, and both tools have real negative tests. Checked the case most likely to be incoherent, whether the notice and the rendered 1970 date in search_applications tell the same story, and they do, the notice explicitly ties the epoch date to never-observed and static uploads rather than contradicting the field.

One thing I looked at and am deliberately not raising, so it does not come up again. The two tools detect the sentinel differently, a null-safe wrapper check in SearchApplicationsTool and bare primitive equality in ListApplicationsByCveTool. That is correct in both places rather than inconsistent, because Application.getLastSeen() returns a Long and App.lastSeen is a primitive long.

Going past the ticket to add the conditional notices and cover search_applications was the right call, the sentinel renders worse there than in the tool the ticket was filed for.

PR #210, opened 2026-08-09
AIML-1239 Add PIT mutation testing to contrast-mcp-core

ChrisEdwards:
**Jira:** [AIML-1239](https://contrast.atlassian.net/browse/AIML-1239)

## Summary
Add PIT mutation testing to contrast-mcp-core so the build catches tests that run code without verifying behavior, a class of defect that line-coverage gates cannot detect.

- Test-strength floor of 91% enforced in CI on every pull request (measured 94%)
- Lombok-generated code excluded from both PIT and JaCoCo via `@lombok.Generated`
- Two FilterHelper boundary mutants killed, two dead-code methods removed

## Why This Change Exists
In November 2025, an always-true assertion (`caughtException || true`) shipped through code review and CI. JaCoCo reported the line as covered because the test executed it. PIT would have flagged it instantly because mutating the production code under that assertion would not flip the test result. This PR wires PIT into the build so that class of problem is caught mechanically going forward.

## Approach
Gate on test strength rather than mutation score. Test strength is killed/(killed+survived), which excludes NO_COVERAGE mutants. That isolates test quality from coverage quantity and avoids double-counting what JaCoCo already enforces. The floor sits a few points under the measured figure, matching the `coverageMinimums` convention.

The Gradle plugin (`info.solidsoft.pitest` 1.19.0) is declared in the root `build.gradle` with `apply false` and applied only in `contrast-mcp-core`. All dependency versions live in `gradle.properties` and pass the 7-day soak window (pitest 1.25.9 was excluded at 4 days old).

### Build infrastructure
- `build.gradle` — pitest plugin declared with `apply false`
- `contrast-mcp-core/build.gradle` — plugin applied, configured with `testStrengthThreshold = 91`, dated baseline comment
- `gradle.properties` — `pitestGradlePluginVersion`, `pitestCoreVersion`, `pitestJunit5PluginVersion`
- `lombok.config` (new) — `lombok.addLombokGeneratedAnnotation = true` and `config.stopBubbling = true`

### CI and developer workflow
- `.github/workflows/build.yml` — mutation testing step on pull requests, PIT reports added to artifact upload
- `Makefile` — `make mutation` target (quiet/verbose pattern)
- `CLAUDE.md` — mutation testing documented in build commands and coverage sections

### Test fixes and dead code removal
- `FilterHelperTest.java` — four boundary tests for `isSupportedTimestampMillis` at `MIN_EPOCH_MILLIS` (0), `MAX_EPOCH_MILLIS` (253402300799999), and one past each boundary
- `FilterHelper.java` — removed `parseDate` and `parseCommaSeparatedLowerCase` (zero callers across all Contrast repos, verified via `gh search code`). Fixed javadoc `@example` references to the removed method. **This is a breaking API change** to the published `contrast-mcp-core` artifact for any external consumer that calls these methods.

## Testing
- `make check-test` passes (1080 tests, 0 failures)
- `make mutation` passes (868 mutants, 792 killed, test strength 94%)
- Gate failure verified: setting `testStrengthThreshold = 99` produces `BUILD FAILED` with `Test strength score of 94 is below threshold of 99`
- Four new boundary tests kill the two `ConditionalsBoundaryMutator` survivors on `FilterHelper.isSupportedTimestampMillis`
- Two pre-existing integration test failures on main (`SearchAttacksToolIT`, `ListApplicationLibrariesToolIT`) are data-dependent and unrelated

- **API break:** `FilterHelper.parseDate` and `parseCommaSeparatedLowerCase` were removed from the published core module. Zero callers exist in any Contrast repo (verified org-wide), but external consumers of `contrast-mcp-core` could break.
- **Equivalent mutants:** 7 of the 52 survivors in `MetadataJsonFilterSpec` are equivalent mutants (error-path `return null` preceded by `invalidEntries.add()`). The mutation is externally unobservable through the public `get()` API. These permanently cap achievable test strength for those sites and cannot be fixed without testing private implementation details.
- **Follow-up:** PIT supports incremental analysis through history files, which could speed up a PR-scoped gate that only re-tests changed code.

[AIML-1239]: https://contrast.atlassian.net/browse/AIML-1239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/FilterHelper.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/FilterHelper.java
@@ -99,10 +99,11 @@ public class FilterHelper {
    * @param dateStr Date string in ISO format or epoch timestamp
    * @param paramName Parameter name for error messages (e.g., "lastSeenAfter")
    * @return ParseResult with Date object and optional validation message
-   * @example parseDate("2025-01-15", "lastSeenAfter") → ParseResult(Date, null)
-   * @example parseDate("1704067200000", "lastSeenAfter") → ParseResult(Date, null)
-   * @example parseDate("invalid", "lastSeenAfter") → ParseResult(null, "Invalid date...")
-   * @example parseDate(null, "lastSeenAfter") → ParseResult(null, null)
+   * @example parseDateWithValidation("2025-01-15", "lastSeenAfter") → ParseResult(Date, null)
+   * @example parseDateWithValidation("1704067200000", "lastSeenAfter") → ParseResult(Date, null)
+   * @example parseDateWithValidation("invalid", "lastSeenAfter") → ParseResult(null, "Invalid
+   *     date...")
+   * @example parseDateWithValidation(null, "lastSeenAfter") → ParseResult(null, null)
    */
   public static ParseResult<Date> parseDateWithValidation(String dateStr, String paramName) {
     if (!StringUtils.hasText(dateStr)) {
@@ -192,14 +193,6 @@ public class FilterHelper {
   }
 
   /**
-   * Parse date string (legacy method for backward compatibility). Use parseDateWithValidation() for
-   * new code to get validation messages.
-   */
-  public static Date parseDate(String dateStr) {
-    return parseDateWithValidation(dateStr, "date").getValue();
-  }
-
-  /**
    * Parse comma-separated list and convert to case-insensitive list. Useful for status, severity,
    * and other case-insensitive filters.
    *
@@ -213,22 +206,6 @@ public class FilterHelper {
       return null;
     }
     return parsed.stream().map(String::toUpperCase).toList();
-  }
-
-  /**
-   * Parse comma-separated list and convert to lowercase list. Useful for vulnerability types and
-   * other lowercase filters.
-   *
-   * @param input Comma-separated string
-   * @return List of trimmed, lowercase strings, or null if input is null/empty
-   * @example parseCommaSeparatedLowerCase("SQL-Injection, XSS") → ["sql-injection", "xss"]
-   */
-  public static List<String> parseCommaSeparatedLowerCase(String input) {
-    List<String> parsed = parseCommaSeparated(input);
-    if (parsed == null) {
-      return null;
-    }
-    return parsed.stream().map(String::toLowerCase).toList();
   }
 
   /**
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/FilterHelper.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/FilterHelper.java
@@ -93,8 +93,9 @@ public class FilterHelper {
 
   /**
    * Parse date string in ISO format (YYYY-MM-DD) or epoch timestamp (milliseconds). Tries epoch
-   * timestamp first, then falls back to ISO date format. Returns validation message if format is
-   * invalid.
+   * timestamp first, then falls back to ISO date format. Epoch timestamps outside the supported
+   * range (1970-01-01 through 9999-12-31 in millis) are rejected. Returns validation message if
+   * format is invalid.
    *
    * @param dateStr Date string in ISO format or epoch timestamp
    * @param paramName Parameter name for error messages (e.g., "lastSeenAfter")
@@ -112,6 +113,9 @@ public class FilterHelper {
     try {
       // Try parsing as epoch timestamp first
       long timestamp = Long.parseLong(dateStr.trim());
+      if (!isSupportedTimestampMillis(timestamp)) {
+        return invalidDateResult(dateStr, paramName);
+      }
       return new ParseResult<>(new Date(timestamp));
     } catch (NumberFormatException e) {
       // Try ISO date format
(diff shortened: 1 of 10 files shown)


contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/FilterHelper.java
```diff
@@ -111,6 +113,9 @@ public static ParseResult<Date> parseDateWithValidation(String dateStr, String p
     try {
       // Try parsing as epoch timestamp first
       long timestamp = Long.parseLong(dateStr.trim());
+      if (!isSupportedTimestampMillis(timestamp)) {
```
Alex-Contrast, line 116: **Medium** — this range check is missing from the ISO branch of the same method, and the new javadoc claims otherwise

**Problem:** This PR adds the supported-range guard to the epoch-timestamp branch, but the ISO fallback eight lines below returns without any equivalent check:

```java
} catch (NumberFormatException e) {
  // Try ISO date format
  try {
    LocalDate localDate = LocalDate.parse(dateStr.trim());
    Date parsed = Date.from(localDate.atStartOfDay(ZoneId.systemDefault()).toInstant());
    return new ParseResult<>(parsed);
```

The javadoc added in this same diff states that "Epoch timestamps outside the supported range (1970-01-01 through 9999-12-31 in millis) are rejected", which is now true for one of the two accepted input formats.

**Impact:** The same instant is accepted or rejected depending only on which format the caller used. Ran the exact expression from the ISO branch against `MIN_EPOCH_MILLIS` (0) and `MAX_EPOCH_MILLIS` (253402300799999):

```
1969-12-31       parses=YES  epochMs=-68400000               inRange=false
1960-01-01       parses=YES  epochMs=-315601200000           inRange=false
-0001-01-01      parses=YES  epochMs=-62198737438000         inRange=false
+10000-01-01     parses=YES  epochMs=253402318800000         inRange=false
2025-01-15       parses=YES  epochMs=1736917200000           inRange=true
```

So any pre-epoch ISO date passes unvalidated. A filter of `lastSeenAfter=1960-01-01` is accepted while the equivalent `lastSeenAfter=-315601200000` is correctly rejected, and a negative `Date` then flows on to the caller. `parseTimestampWithValidation` does not have this gap, it checks at both of its parse sites, so this is an asymmetry within `parseDateWithValidation` alone.

**Fix:** Check before returning, mirroring what the epoch branch now does:

```java
Date parsed = Date.from(localDate.atStartOfDay(ZoneId.systemDefault()).toInstant());
if (!isSupportedTimestampMillis(parsed.getTime())) {
  return invalidDateResult(dateStr, paramName);
}
return new ParseResult<>(parsed);
```

Worth a test at both ends given what this PR is about. This is precisely the shape of gap PIT is being introduced to catch, a branch that no test distinguishes, so it would be a good demonstration of the gate earning its place.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/FilterHelperTest.java
```diff
@@ -46,6 +47,11 @@ class FilterHelperTest {
           .toInstant(ZoneOffset.UTC)
           .toEpochMilli(); // 1500000000000L
 
+  // Must stay in sync with FilterHelper's private MIN_EPOCH_MILLIS / MAX_EPOCH_MILLIS.
+  private static final long MIN_EPOCH_MILLIS = 0L;
+  private static final long MAX_EPOCH_MILLIS = 253402300799999L;
```
Alex-Contrast, line 52: **Nit** — boundary constants duplicated from production, non-blocking

**Problem:** `MIN_EPOCH_MILLIS` and `MAX_EPOCH_MILLIS` are redeclared here with literal values that also live in `FilterHelper` as `private static final`.

**Impact:** Small but real drift risk. If the production range ever changes, these boundary tests keep passing against the old boundary and quietly stop testing the boundary they are named for. That is a slightly awkward failure mode in a PR whose subject is catching tests that pass without verifying behaviour.

**Fix:** Make the two constants in `FilterHelper` package-private and reference them directly. Package-private is not published API, so it does not interact with the artifact-surface question elsewhere in this PR.

Alex-Contrast (commented):
Reviewed the PIT mutation testing addition. Two inline comments, one Medium and one nit. The infrastructure work itself looks solid.

**Things I checked and am not raising:**

- The gate genuinely fails CI. This was the claim most worth verifying, since a mutation gate that reports without failing is the usual way this lands ineffective. `.github/workflows/build.yml` runs `:contrast-mcp-core:pitest` on `pull_request` with no `continue-on-error`, and it reads the `testStrengthThreshold = 91` from `contrast-mcp-core/build.gradle`. It gates.
- The four new `FilterHelperTest` boundary tests actually observe the mutation rather than merely executing the boundary. Flipping `>=` to `>` at min and `<=` to `<` at max each break a specific new assertion.
- Nothing was deleted alongside the two removed methods in a way that would inflate the measured figure.
- Gating on test strength rather than mutation score is the right call, and the reasoning in the ticket about not double-counting what JaCoCo already enforces holds up.

**Two things worth a sentence, neither a code finding:**

The removal of `FilterHelper.parseDate` and `parseCommaSeparatedLowerCase` is a breaking change to the published `contrast-mcp-core` artifact, and `verifyCorePublicationMetadata` would not have caught it, it validates POM and module metadata dependencies rather than class member signatures. The zero-caller search justifies the deletion for internal consumers, but `RELEASING.md` requires a MAJOR release for a breaking public API change and the PR does not mention the version implication. Mostly a release-sequencing question for you rather than something to change here.

Excluding Lombok-generated code from JaCoCo as well as PIT makes the existing coverage floors easier to pass rather than harder, since it shrinks the denominator. No floor value was lowered so there is nothing wrong with it, but it does mean the coverage numbers before and after this PR are not directly comparable. Worth knowing if anyone reads the delta as an improvement.

Produced with model assistance. Both findings were verified against the source on this branch, and the range-check behaviour in the Medium was confirmed by execution rather than by reading.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/FilterHelper.java
ChrisEdwards: Fixed in 0954bfb on #211. Added `isSupportedTimestampMillis` check to the ISO branch and four boundary tests (pre-epoch, far-future, valid mid-range, epoch start).

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/base/FilterHelperTest.java
ChrisEdwards: Fixed in 0954bfb on #211. Constants are now package-private, tests reference `FilterHelper.MIN_EPOCH_MILLIS` and `FilterHelper.MAX_EPOCH_MILLIS` directly.

ChrisEdwards:
Re: the breaking-API-change observation on `FilterHelper.parseDate` and `parseCommaSeparatedLowerCase` — noted. We are the sole consumer of `contrast-mcp-core` and I verified both methods are unused across Contrast repos, so this does not need a major version bump.

The Lombok/JaCoCo denominator note is fair. The floors stayed constant, so no coverage was lost.

Alex-Contrast (approved):

PR #211, opened 2026-08-09
AIML-1239 Fix PIT mutation survivors

ChrisEdwards:
**Jira:** [AIML-1239](https://contrast.atlassian.net/browse/AIML-1239)

## Summary

Strengthen tests to kill real PIT mutation survivors without changing production code. Equivalent mutants in well-structured helper methods are accepted as a natural cost of clean decomposition, not a problem to fix by restructuring.

- Test strength measured at 94.5%, floor raised from 91% to 92%
- 3 test files strengthened, 0 production logic changes
- Equivalent mutants in `HttpRequestRedactor` helpers and `GetVulnerabilityTool.extractStackTraces` accommodated by threshold headroom

## Why This Change Exists

The parent PR (#210) added PIT mutation testing and set an initial test-strength floor of 91%. Some survivors indicated genuinely weak test assertions (e.g. mocking the mapper instead of exercising it, missing boundary tests). Others were equivalent mutants in small helper methods where PIT can mutate a branch that produces identical output from the public API. This branch fixes the genuine gaps without restructuring production code to eliminate equivalent-mutant sites.

## Approach

Strengthen tests to observe real mutations. Accept equivalent mutants and set the floor with headroom to accommodate them, matching the convention already used for JaCoCo coverage floors.

Specifically rejected the alternative of collapsing helpers into regex or Optional chains to eliminate mutant sites. That approach trades readability for a metric, the opposite of what mutation testing should encourage.

### Build config
- `contrast-mcp-core/build.gradle` — Raised `testStrengthThreshold` from 91 to 92, updated measured-baseline comment (882 mutants, 812 killed, 47 survived including equivalent mutants)

### Tests
- `GetVulnerabilityToolTest.java` — Main enrichment test uses real `VulnerabilityMapper` instead of a mock, asserts actual `Vulnerability` fields (recommendation, HTTP request content, stack traces, libraries); added test for null recommendation; strengthened event summary test data with nulls and non-trigger events to exercise filter paths
- `CursorPaginationParamsTest.java` — Added boundary tests for page size 1 and `MAX_PAGE_SIZE`
- `HttpRequestRedactorTest.java` — Added edge-case tests for colon-first-character lines, spaces in header names, and single-character header names

## Testing

All tests pass (`make check-test`, 1092 tests, 0 failures). Mutation testing passes (`make mutation`, test strength 94.5% vs floor of 92%). The new tests kill their target mutants: the real-mapper test catches `EmptyObjectReturnValsMutator` survivors in mapper lambdas, boundary page-size tests catch `ConditionalsBoundaryMutator` survivors, and the redactor edge-case tests cover previously untested input shapes.

Remaining survivors (47) are either equivalent mutants in helper methods (e.g. `extractStackTraces` returning empty list vs new empty ArrayList when event is null) or no-coverage sites already gated by JaCoCo floors.

[AIML-1239]: https://contrast.atlassian.net/browse/AIML-1239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alex-Contrast (approved):

A PR whose purpose is to make a mutation gate pass is exactly where a test gets quietly weakened, so I checked that specifically rather than taking the description on trust. It holds. No file under src/main is touched, and the assertions this diff removes were all captor-based checks against a mocked VulnerabilityMapper, replaced by exact-value assertions on the real mapper's output (isEqualTo, containsExactly, isNull). isSameAs disappears legitimately because there is no longer a mock instance to be identical to. Net movement is from asserting mapper inputs to asserting mapped outputs, which is strictly stronger and is the point of the change.

Agree with accepting the equivalent mutants rather than restructuring helpers to eliminate mutant sites. Trading readability for a metric would defeat the purpose.

One sequencing note, not a blocker. This sits on #210, where I have left a Medium about the ISO branch of parseDateWithValidation missing its range check. If that gets fixed in #210 this branch rebases and this approval will be against a commit that no longer exists, so treat it as approval of the delta rather than of a specific SHA.

PR #212, opened 2026-08-10
AIML-1349 Release v2.4.0 changelog and error message fix

ChrisEdwards:
## Summary

- Bring `[Unreleased]` in CHANGELOG.md up to date with all user-facing changes merged since v2.3.0, stamped as v2.4.0
- Fix the `list_applications_by_cve` HTTP 500 error message, which referenced `search_cves` and `get_cve_impact`. Those tools exist only in the hosted MCP server (aiml-services). The `contrast-mcp-core` library is shared between both servers, and the stdio server has no access to hosted-only tools. An agent receiving this error from the stdio server would try to call nonexistent tools and fail. The message also referenced "NorthStar CVE Shield data," which is a hosted-server concept that does not apply to stdio users connecting to TeamServer Classic. The corrected message explains the SCA library data limitation without referencing tools or concepts that may not be available to the caller.

## Test plan

- [x] Changelog follows house style
- [x] Only user-facing changes are documented
- [x] Error message no longer references hosted-server tools
- [x] `make check-test` passes (1100 tests, 0 failures)
- [x] Changed-file coverage gate passes

### DIFF
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
@@ -52,9 +52,9 @@ public class ListApplicationsByCveTool extends SingleTool<ListApplicationsByCveP
 
   private static final int HTTP_INTERNAL_SERVER_ERROR = 500;
   private static final String INTERNAL_SERVER_ERROR_MESSAGE =
-      "The service returned an error. This happens for CVEs the SCA library data does not"
-          + " recognize, including CVEs that exist only in NorthStar CVE Shield data. Verify the"
-          + " CVE with search_cves or get_cve_impact, or retry later if the service is failing.";
+      "The service returned an error. This typically happens for CVEs that the SCA library data"
+          + " does not recognize. Verify the CVE identifier is correct and retry later if the"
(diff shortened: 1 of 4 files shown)


contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/library/ListApplicationsByCveTool.java
```diff
@@ -52,9 +52,9 @@ public class ListApplicationsByCveTool extends SingleTool<ListApplicationsByCveP
 
   private static final int HTTP_INTERNAL_SERVER_ERROR = 500;
   private static final String INTERNAL_SERVER_ERROR_MESSAGE =
-      "The service returned an error. This happens for CVEs the SCA library data does not"
-          + " recognize, including CVEs that exist only in NorthStar CVE Shield data. Verify the"
-          + " CVE with search_cves or get_cve_impact, or retry later if the service is failing.";
+      "The service returned an error. This typically happens for CVEs that the SCA library data"
+          + " does not recognize. Verify the CVE identifier is correct and retry later if the"
```
ChrisEdwards, line 56: Had to remove these tool names at the last second as they don't exist in the stdio MCP server. A rule exists for that, but its in MCP_STANDARDS.md, so I made that clearer in CLAUDE.md as well.

dougj-contrast (approved):

PR #217, opened 2026-08-11
AIML-1353 Validate vulnTypes against live org catalogue

ChrisEdwards:
**Jira:** [AIML-1353](https://contrast.atlassian.net/browse/AIML-1353)

## Summary

Both `search_vulnerabilities` and `search_app_vulnerabilities` now validate the `vulnTypes` parameter against the live organization rule catalogue before searching. Invalid values produce an explicit error message pointing the caller to `list_vulnerability_types`, instead of silently returning empty results that look identical to a legitimate "no vulnerabilities found" response.

- Invalid or misspelled vulnerability types are caught before the search API is called
- Values are matched case-insensitively and whitespace-trimmed, with the canonical form forwarded downstream
- The fail-closed design refuses the request when the rule catalogue is unavailable

## Why This Change Exists

An AI agent that misspells a vulnerability type (for example `sqli` instead of `sql-injection`) gets back `success=true, items=[], totalItems=0`. That response is indistinguishable from a genuine empty result, so the agent concludes there are no matching vulnerabilities and moves on. The `severities`, `statuses`, and `environments` parameters already reject invalid values with clear error messages. `vulnTypes` was the inconsistent outlier.

## Approach

A new `VulnerabilityTypeValidator` service fetches the org's rules via `ContrastApiClient.getRules()`, builds a case-insensitive lookup map, and returns the canonical spelling for each supplied value. Any unrecognized values cause a single `IllegalArgumentException` that names every bad value and points to `list_vulnerability_types`. The `PaginatedTool` base class already converts `IllegalArgumentException` to a `success=false` response, so no error-handling plumbing was needed.

Validation happens in `doExecute()` of both tools, not in the params layer, because the params layer has no access to the API client. This matches the precedent set by session-metadata filter resolution in `SearchAppVulnerabilitiesTool`.

Rejected alternatives and the reasoning behind them are documented in the bead's design field.

### Validator
- `VulnerabilityTypeValidator.java` (new) - shared service injected into both search tools

### Search tools
- `SearchVulnerabilitiesTool.java` - calls the validator at the top of `doExecute()` when `vulnTypes` is present, overwrites the filter body with the canonical list
- `SearchAppVulnerabilitiesTool.java` - same validation wired in the same position

### Standards
- `MCP_STANDARDS.md` - new "Enumerated String Parameters" section codifying the Postel rule for all enumerated string params
- `CONTEXT.md` - added "Vulnerability type" to the domain glossary

### Tests
- `VulnerabilityTypeValidatorTest.java` (new) - 8 tests covering valid values, case/whitespace normalization, mixed valid+invalid rejection, null/empty catalogue fail-closed, and exception propagation
- `SearchVulnerabilitiesToolTest.java` - 2 new tests for invalid and mixed vulnTypes rejection, updated existing routing test to exercise case normalization
- `SearchAppVulnerabilitiesToolTest.java` - 3 new tests for invalid, mixed, and canonical-forwarding behavior
- `VulnerabilityFilterParamsTest.java` - updated section header comment to reflect that vulnTypes is now validated at execution time
- `SearchVulnerabilitiesLocalParityTest.java`, `SearchAppVulnerabilitiesLocalParityTest.java` - updated constructors for new validator dependency

## Testing

The validator has dedicated unit tests for every path: valid input, case normalization, whitespace trimming, single and multiple invalid values, mixed valid+invalid lists, null catalogue, empty catalogue, blank-named-only catalogue, and API failure propagation. Both tool tests verify that invalid vulnTypes produce `success=false` with the error constant, that the search API is never called, and that valid values with non-canonical casing are forwarded in canonical form. All 1,133 unit tests pass, coverage floors and PIT mutation strength floors are met.

Integration test coverage now verifies the rejection path and exact validation error shape, matching the `SearchAttacksToolIT` pattern.

[AIML-1353]: https://contrast.atlassian.net/browse/AIML-1353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -20,6 +20,10 @@ _Avoid_: conflating SmartFix with the AutoRemediated status or with Auto-Verifie
 The NorthStar status vocabulary on Issue entities. The values are REPORTED, CONFIRMED, REMEDIATED, NOT_A_PROBLEM, and FIXED. Suspicious and AutoRemediated exist only as Vulnerability statuses, never as Issue Statuses.
 _Avoid_: using this term for legacy Vulnerability statuses
 
+**Vulnerability type**:
+The kind of vulnerability a finding represents, for example `sql-injection` or `reflected-xss`. Values are lowercase-hyphenated and org-specific, discoverable via `list_vulnerability_types`. Synonymous with the TeamServer term "rule name", a historical artifact: the Contrast Agent used a detection rule per vulnerability type, so the `sql-injection` rule found SQL injection vulnerabilities. Vulnerability type is the canonical term.
+_Avoid_: rule name, rule type, except when naming the raw TeamServer API surface (e.g. `getRules()`)
+
 **Status display label**:
 The human-readable status text TeamServer returns on vulnerability records, for example "Remediated - Auto-Verified" for a vulnerability in the AutoRemediated status. Filters take canonical Vulnerability statuses, and this server's tool results carry canonical Vulnerability statuses, not display labels.
 _Avoid_: treating a display label as a valid filter value or exposing one in a tool result
--- a/MCP_STANDARDS.md
+++ b/MCP_STANDARDS.md
@@ -121,6 +121,10 @@ Tools returning analytical data (reports, coverage, metadata) may use `get_*` ev
 - **Plural** for comma-separated: `severities`, `statuses`, `environments`
 - **Singular** for single values: `appId`, `keyword`, `sort`
 
+### Enumerated String Parameters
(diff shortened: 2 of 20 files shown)


ChrisEdwards:
Dissolving native stack to rebuild on correct root.

ChrisEdwards:
Recreating to join unified native stack.

PR #219, opened 2026-08-11
AIML-1354 Explain route coverage license denials

ChrisEdwards:
**Jira:** [AIML-1354](https://contrast.atlassian.net/browse/AIML-1354)

## Summary

When an AI agent calls `get_route_coverage` on an unlicensed or archived application, TeamServer returns an opaque HTTP 403 ("Authorization failure") that the MCP server surfaces as "Access denied or resource not found. Verify credentials and that the resource ID is correct." The agent then retries in a loop, questions its credentials, and reports a broken setup. This PR discriminates the cause of the 403 and returns a truthful, actionable error that tells the agent exactly what is wrong and what to do about it.

- Unlicensed, archived, and residual-denial cases each get a distinct, non-retryable error message
- Zero added API calls on the happy path
- Shared `ActionableToolErrorException` mechanism for future use by sibling ticket AIML-1137

## Why This Change Exists

TeamServer's `RouteCoverageAnalyticsRestController` enforces `isAppEnterprise(#app)` at the class level, which requires the application's merged license service level to be exactly "Licensed". Any other level produces an undiscriminated 403. The MCP server's generic 403 handler cannot distinguish a credential problem from a license denial, so it emits advice that is factually wrong for the license case. Verified against staging: 9/9 unlicensed apps across six languages returned 403, 2/2 licensed apps returned 200, and an app with no coverage data returns 200 with an empty routes list.

## Approach

Follows the AIML-1137 precedent: discriminate only after a 403 arrives, never speculatively. On a 403 from the route coverage endpoint, `GetRouteCoverageTool.fetchRouteCoverage` catches the `UnauthorizedException`, calls `ApplicationLicenseDiscriminator.discriminate`, and throws an `ActionableToolErrorException` with the appropriate message. The discriminator fetches the single application with license expansion (an endpoint that remains available for archived and unlicensed apps) and classifies the state as ARCHIVED, UNLICENSED, or LICENSED. If the discriminator lookup itself fails, the original 403 is rethrown so the generic message surfaces, which is then actually true.

`ActionableToolErrorException` is a new exception type caught by `SingleTool.executePipeline` and surfaced verbatim as the error message. This replaces the need for `IllegalArgumentException` overloading and will be reused by AIML-1137.

### Core domain (`contrast-mcp-core`)
- `ApplicationLicenseDiscriminator` (new) — reusable service that fetches an application's license state after a 403 and classifies it as ARCHIVED, UNLICENSED, or LICENSED
- `ApplicationLicense`, `ApplicationResponse` (new) — data models for the single-application-with-license response
- `Application` — added `license` field with `@SerializedName`
- `ActionableToolErrorException` (new) — exception whose message is safe and actionable for the caller, caught by `SingleTool`
- `SingleTool` — catches `ActionableToolErrorException` and surfaces its message verbatim
- `GetRouteCoverageTool` — extracted `fetchRouteCoverage` that intercepts 403s, runs the discriminator, and throws the appropriate actionable error; updated tool description to mention the Assess license requirement
- `ContrastApiClient` — added `getApplicationWithLicense(appId)` method

### Local app (`contrast-mcp-stdio-app`)
- `SdkApiClient` — implements `getApplicationWithLicense` via `SDKExtension`
- `SDKExtension` — added `getApplicationWithLicense` using the single-application endpoint with `expand=license`; also added `license` to the existing `getApplications` expansion list

### Domain glossary
- `CONTEXT.md` — added "Licensed application" and "Route coverage" terms

## Testing

**Unit tests:**
- `ApplicationLicenseDiscriminatorTest` — all discriminator branches: archived app, unlicensed app, null license, licensed app, fetch failure (rethrows original), null application (rethrows original)
- `GetRouteCoverageToolTest` — three new tests covering the unlicensed, archived, and residual-denial 403 paths; existing happy-path test verifies discriminator is never invoked on success
- `SingleToolTest` — verifies `ActionableToolErrorException` message surfaces verbatim as user error
- `SDKExtensionTest` — verifies `getApplicationWithLicense` sends the correct URL with `expand=license`
- `SDKExtensionResponseHandlingTest` — updated to verify the applications list endpoint now includes `license` in its expansion

**Parity test:**
- `GetRouteCoverageLocalParityTest` — new test wires the real `SdkApiClient` → `SDKExtension` → discriminator chain end-to-end with mocks and verifies the unlicensed error surfaces through the local client

**Integration test:**
- `GetRouteCoverageToolIT` — new test discovers an unlicensed application via the license expansion on `search_applications` results, calls `get_route_coverage` against it, and asserts the exact `UNLICENSED_APPLICATION_ERROR` message. Discovery version bumped to invalidate stale cached test data.

- Error message wording ships without blocking product sign-off. The messages are factual and actionable, but phrasing may be refined after review.
- The discriminator makes one extra API call only on the 403 path. Happy-path performance is unchanged.
- Follow-up bead `mcp-q5zb8` tracks surfacing license level in `search_applications` results so agents can predict which apps will work before calling `get_route_coverage`.
- Sibling ticket AIML-1137 will reuse `ApplicationLicenseDiscriminator` and `ActionableToolErrorException` for vulnerability detail reads.
- Hosted MCP (`aiml-services`) needs a corresponding `getApplicationWithLicense` implementation, tracked separately.

[AIML-1354]: https://contrast.atlassian.net/browse/AIML-1354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -480,6 +480,12 @@ Run the harness-engineering playbooks against this repo on demand via `/harness-
 
 `/update-changelog` brings `[Unreleased]` in CHANGELOG.md up to date against everything merged since the last release tag, audits completeness with a read-only subagent per range, and lands the result on a branch after user approval. Run before dispatching the Gradle Release workflow or when explicitly asked. See `.claude/skills/update-changelog/`.
 
+## Spawning Codex via Herdr
+
+When spawning a Codex agent through Herdr:
+
+1. `herdr agent start <name> --kind codex --pane <pane-id>` with no extra args after `--`. Default Codex startup is correct. Do not pass `--full-auto` or `--permission-mode auto`.
+2. `herdr agent prompt <name> "<brief>" --wait` with NO `--timeout` flag. Codex runs often exceed 10 minutes. Omitting `--timeout` lets Herdr wait indefinitely. The Bash tool's own 10-minute ceiling applies, so run the prompt call with `run_in_background: true` and wait for the task notification.
 
 
 @SECURITY.md
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -24,6 +24,14 @@ _Avoid_: using this term for legacy Vulnerability statuses
 The kind of vulnerability a finding represents, for example `sql-injection` or `reflected-xss`. Values are lowercase-hyphenated and org-specific, discoverable via `list_vulnerability_types`. Synonymous with the TeamServer term "rule name", a historical artifact: the Contrast Agent used a detection rule per vulnerability type, so the `sql-injection` rule found SQL injection vulnerabilities. Vulnerability type is the canonical term.
 _Avoid_: rule name, rule type, except when naming the raw TeamServer API surface (e.g. `getRules()`)
 
+**Licensed application**:
+An application whose merged license service level in TeamServer is exactly Licensed, meaning it holds an Assess license. TeamServer refuses route coverage and single-vulnerability detail reads for unlicensed applications with an opaque HTTP 403 "Authorization failure".
+_Avoid_: describing those 403s as credential problems, missing data, or language limitations
+
+**Route coverage**:
+The set of routes an Assess agent observed in an application, each either discovered (found in the code) or exercised (hit by an HTTP request).
+_Avoid_: route data, endpoint coverage
+
 **Status display label**:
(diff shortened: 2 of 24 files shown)


ChrisEdwards:
Dissolving native stack to rebuild on correct root.

ChrisEdwards:
Recreating to join unified native stack.

PR #220, opened 2026-08-11
AIML-1355 Onboard ai-sdlc plugin to repo

ChrisEdwards:
**Jira:** [AIML-1355](https://contrast.atlassian.net/browse/AIML-1355)

## Summary
Onboard the ai-sdlc Claude plugin to this repo by consolidating agent instruction files, adding missing architecture documentation, and strengthening the pre-commit hook with secret scanning and shell linting.

- Single canonical agent instructions file with symlinks for cross-tool compatibility
- Pre-commit hook now runs gitleaks, shellcheck, and end-of-file-fixer before Java checks
- REPO-CONTEXT.md created with planning checklist for future design reviews

## Why This Change Exists
- Multiple agent instruction files (AGENTS.md, CLAUDE.md) diverged over time, with overlapping and sometimes contradictory content
- The pre-commit hook only covered Java-specific checks (spotless, checkstyle, tests) while a `.pre-commit-config.yaml` defined gitleaks and shellcheck but was never installed
- No REPO-CONTEXT.md existed for the ai-sdlc `plan` skill to read during design reviews

### Agent instruction files
- `CLAUDE.md` — added communication style, REPO-CONTEXT.md reference, module split (contrast-mcp-core vs contrast-mcp-stdio-app with composite-build instructions), and ast-grep refactoring section
- `AGENTS.md` — replaced with a symlink to CLAUDE.md (unique content merged into CLAUDE.md first)
- `.github/copilot-instructions.md` — new symlink to CLAUDE.md for GitHub Copilot compatibility

### Repo context
- `REPO-CONTEXT.md` — new file with running/testing instructions and a planning checklist (cross-component data-contract changes, integration test updates)

### Pre-commit hook
- `scripts/git-hooks/pre-commit` — added gitleaks secret scanning (hard fail), end-of-file newline fixer, and shellcheck linting before the existing Java checks. Tools that are not installed warn but do not block.

## Testing
- Ran `make install-hooks` to install the updated hook
- Verified the hook passes on a clean tree (gitleaks warns when not installed, shellcheck runs, Java checks skipped when no .java files staged)
- Committed successfully with the new hook active: gitleaks scanned and found no leaks, shellcheck passed
- Verified all three symlinks resolve to the same inode via `stat -L -f %i`

[AIML-1355]: https://contrast.atlassian.net/browse/AIML-1355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/git-hooks/pre-commit
```diff
@@ -1,17 +1,58 @@
...
+    echo "Commit blocked: gitleaks found potential secrets in staged changes."
+    exit 1
+  fi
+else
+  echo "Warning: gitleaks not installed, skipping secret scan (brew install gitleaks)"
+fi
+
+# --- end-of-file-fixer: ensure staged files end with a newline ---
+while IFS= read -r -d '' file; do
+  if [ -s "$file" ] && [ "$(tail -c 1 "$file" | wc -l)" -eq 0 ]; then
```
Alex-Contrast, line 23: **High** — the end-of-file fixer corrupts binary files.

**Problem.** This loop iterates every staged file from `--diff-filter=ACM` with no text/binary test. For a binary whose final byte is not `0x0A`, `tail -c 1 | wc -l` yields `0`, so `printf '\n' >> "$file"` appends a byte into the binary and the following `git add` stages the corruption.

**Impact.** Reproduced with the exact loop from this file. A 33-byte PNG (signature + IHDR, final byte `0x89`), staged, then run through it:

```
before: 33 bytes  4ffd8bb30991e3a6f28d1d03f1aedcd02ccf8e0cc16bb9e969e7bc2bda1ddf03
after : 34 bytes  31763180e969324abfa5db799db8bc2e90bfaa27fe3c71e2e98a367b617b6022
RESULT: BINARY CORRUPTED (+1 byte appended)
```

This repo tracks 14 binaries, including `gradle/wrapper/gradle-wrapper.jar`, `images/*.png`, and `images/intellij.gif`. Any commit that stages one corrupts it.

**Fix.** A `grep -I` guard is the smallest change and needs no new dependency, since `grep -I` treats a binary file as non-matching:

```bash
while IFS= read -r -d '' file; do
  [ -s "$file" ] || continue
  grep -Iq . "$file" || continue   # skip binaries
  if [ "$(tail -c 1 "$file" | wc -l)" -eq 0 ]; then
    ...
```

The existing `[ -s "$file" ]` already excludes empty files, so `grep -Iq .` cannot false-negative on one.

Alex-Contrast (commented):
Cross-model review of the ai-sdlc onboarding change, triaged against the branch before posting. Three findings, all in `scripts/git-hooks/pre-commit` or the `AGENTS.md` consolidation. The High and the Medium were both reproduced against the loop as written rather than reasoned about, and the repro output is in each comment.

The Medium is worth reading even if you disagree with the High, because the two interact. The eof-fixer stages content after gitleaks has already scanned, so the sweep is also a way for unscanned content to reach a commit.

**Considered and deliberately not raised**, so you do not have to re-litigate them:

- **shellcheck shebang detection.** Reported by the model reviewer as missing `.sh` files without a shebang. Rejected. All four tracked `.sh` files have shebangs, and shebang detection also catches this repo's extension-less scripts (`gradlew`, `scripts/ralph*`, `scripts/git-hooks/pre-commit` itself), which extension matching would miss. The current rule is the better one.
- **`Testing Anti-Patterns` as a third dropped section.** A heading-level diff flags it as missing, but `CLAUDE.md:325` supersedes it with a far broader list. Not a loss.
- **Line 62, `git diff --name-only -z -- '*.java' | xargs -r -0 git add`.** Same whole-file sweep shape as the Medium. Not raised as a finding because I did not establish whether this PR introduced it, but it is worth a look while you are in this file.

**Open questions**, genuine questions rather than change requests in disguise:

1. Should the dropped `history/` planning-doc rule and the Multi-Agent Collaboration guidance be restored to `CLAUDE.md`, or were they dropped on purpose?
2. For machines without gitleaks installed, is a warning the right call for local development, with CI as the real gate? The PR description calls gitleaks a hard fail, which is true only when the binary is present.

Produced with model assistance and verified against the branch by hand.

scripts/git-hooks/pre-commit
```diff
@@ -1,17 +1,58 @@
...
+    printf '\n' >> "$file"
+    git add -- "$file"
```
Alex-Contrast, line 25: **Medium** — this stages unrelated unstaged work, and it does so after gitleaks has already run.

**Problem.** `git add -- "$file"` stages the whole working-tree file. A developer who staged selectively with `git add -p` gets their remaining unstaged edits in that file committed too, whenever the file happens to lack a trailing newline.

**Impact.** Two layers. The first is the surprise itself, reproduced by staging one hunk, leaving another edit unstaged, and running the loop:

```
=== staged AFTER hook ===
 -line one
 +line one CHANGED
  line two
 +SECRET=hunter2
RESULT: unstaged line SWEPT INTO COMMIT
```

The second is the one that makes this more than an annoyance. gitleaks scans `--staged` at line 12 of this file. This line stages *new* content at line 25, after that scan has finished. Anything swept in is committed without ever being scanned for secrets, which is precisely what the repro above demonstrates. That is a gap in the secret-scanning gate this PR is adding, not just a staging surprise.

**Fix.** Only auto-fix when the file has nothing unstaged, which also preserves the scan's guarantee:

```bash
if git diff --quiet -- "$file"; then
  printf '\n' >> "$file"
  git add -- "$file"
else
  echo "Warning: $file has no trailing newline but has unstaged changes; fix it manually."
fi
```

AGENTS.md
```diff
@@ -1,204 +0,0 @@
...
-  - If creating Jira ticket for existing bead, update title: `br update <id> --title "AIML-XXX: ..."`
-- **Set the `external_ref` to the Jira issue ID**
-  - `br update <id> --external-ref AIML-224`
-- This creates bidirectional linkage:
-  - Bead tracks local technical work (code, tests, implementation)
-  - Jira tracks project management (planning, stakeholders, timeline)
-  - Clear connection between both via issue ID in title and external_ref field
-
-
-### Managing AI-Generated Planning Documents
```
Alex-Contrast, line 122: **Low** — two sections were dropped rather than merged when this file became a symlink.

**Problem.** The PR description states that unique `AGENTS.md` content was merged into `CLAUDE.md` before the symlink replaced it. Two sections were not, and their substance survives nowhere on this branch:

- **Managing AI-Generated Planning Documents**, the rule to store ephemeral planning docs in a `history/` directory. The string `history/` appears in neither `CLAUDE.md` nor `REPO-CONTEXT.md`.
- **Multi-Agent Collaboration**, including "do NOT revert unexpected changes without asking" and the guidance to check active bead status before editing shared files. Neither `Multi-Agent` nor `revert unexpected` appears anywhere on the branch.

**Impact.** Both are agent-behaviour rules, so their absence is silent. Nothing fails, an agent just stops following them. The `do not revert another agent's uncommitted work` rule is the costlier of the two to lose.

**Fix.** Fold both into `CLAUDE.md`, or state in the PR description that they were dropped deliberately so the record is accurate.

For what it is worth, `Testing Anti-Patterns` also disappears as a heading but is genuinely superseded, the new block at `CLAUDE.md:325` is broader than the two bullets it replaced. No action needed there.

ChrisEdwards: This was intentional.

scripts/git-hooks/pre-commit
ChrisEdwards: Fixed in ca1568f on the tip branch (AIML-1354-route-coverage-license-403). Added `grep -Iq . "$file" || continue` after the `[ -s ]` check, so binaries skip the newline fixer entirely.

scripts/git-hooks/pre-commit
ChrisEdwards: Fixed in ca1568f on the tip branch (AIML-1354-route-coverage-license-403). Two changes:

1. The EOF fixer loop now guards with `git diff --quiet -- "$file"` — files with unstaged changes get a warning instead of being staged.
2. Gitleaks moved to run after all staging operations. Extracted into a `scan_staged_secrets()` function called in both the no-Java early-exit path and after `spotlessApply` + `git add`, so it always scans the final staged content.

ChrisEdwards:
Re the open questions from the review:

1. **Dropped AGENTS.md sections** — intentional, as noted in the resolved thread.

2. **gitleaks as warning when not installed** — the local hook is a soft gate by design: gitleaks is recommended but not required locally. We looked at adding gitleaks to CI as a hard gate, but `gitleaks-action` [requires a paid license](https://github.com/Contrast-Security-OSS/mcp-contrast/actions/runs/31507041394/job/93831159355) for CI use. The local pre-commit hook remains the only gitleaks gate, so the warning-when-absent behavior is correct.

Alex-Contrast (approved):

PR #223, opened 2026-08-11
AIML-1353 Validate vulnTypes against live org catalogue

ChrisEdwards:
**Jira:** [AIML-1353](https://contrast.atlassian.net/browse/AIML-1353)

## Summary

Both `search_vulnerabilities` and `search_app_vulnerabilities` now validate the `vulnTypes` parameter against the live organization rule catalogue before searching. Invalid values produce an explicit error message pointing the caller to `list_vulnerability_types`, instead of silently returning empty results that look identical to a legitimate "no vulnerabilities found" response.

- Invalid or misspelled vulnerability types are caught before the search API is called
- Values are matched case-insensitively and whitespace-trimmed, with the canonical form forwarded downstream
- The fail-closed design refuses the request when the rule catalogue is unavailable

## Why This Change Exists

An AI agent that misspells a vulnerability type (for example `sqli` instead of `sql-injection`) gets back `success=true, items=[], totalItems=0`. That response is indistinguishable from a genuine empty result, so the agent concludes there are no matching vulnerabilities and moves on. The `severities`, `statuses`, and `environments` parameters already reject invalid values with clear error messages. `vulnTypes` was the inconsistent outlier.

## Approach

A new `VulnerabilityTypeValidator` service fetches the org's rules via `ContrastApiClient.getRules()`, builds a case-insensitive lookup map, and returns the canonical spelling for each supplied value. Any unrecognized values cause a single `IllegalArgumentException` that names every bad value and points to `list_vulnerability_types`. The `PaginatedTool` base class already converts `IllegalArgumentException` to a `success=false` response, so no error-handling plumbing was needed.

Validation happens in `doExecute()` of both tools, not in the params layer, because the params layer has no access to the API client. This matches the precedent set by session-metadata filter resolution in `SearchAppVulnerabilitiesTool`.

Rejected alternatives and the reasoning behind them are documented in the bead's design field.

### Validator
- `VulnerabilityTypeValidator.java` (new) - shared service injected into both search tools

### Search tools
- `SearchVulnerabilitiesTool.java` - calls the validator at the top of `doExecute()` when `vulnTypes` is present, overwrites the filter body with the canonical list
- `SearchAppVulnerabilitiesTool.java` - same validation wired in the same position

### Standards
- `MCP_STANDARDS.md` - new "Enumerated String Parameters" section codifying the Postel rule for all enumerated string params
- `CONTEXT.md` - added "Vulnerability type" to the domain glossary

### Tests
- `VulnerabilityTypeValidatorTest.java` (new) - 8 tests covering valid values, case/whitespace normalization, mixed valid+invalid rejection, null/empty catalogue fail-closed, and exception propagation
- `SearchVulnerabilitiesToolTest.java` - 2 new tests for invalid and mixed vulnTypes rejection, updated existing routing test to exercise case normalization
- `SearchAppVulnerabilitiesToolTest.java` - 3 new tests for invalid, mixed, and canonical-forwarding behavior
- `VulnerabilityFilterParamsTest.java` - updated section header comment to reflect that vulnTypes is now validated at execution time
- `SearchVulnerabilitiesLocalParityTest.java`, `SearchAppVulnerabilitiesLocalParityTest.java` - updated constructors for new validator dependency

## Testing

The validator has dedicated unit tests for every path: valid input, case normalization, whitespace trimming, single and multiple invalid values, mixed valid+invalid lists, null catalogue, empty catalogue, blank-named-only catalogue, and API failure propagation. Both tool tests verify that invalid vulnTypes produce `success=false` with the error constant, that the search API is never called, and that valid values with non-canonical casing are forwarded in canonical form. All 1,133 unit tests pass, coverage floors and PIT mutation strength floors are met.

Integration test coverage now verifies the rejection path and exact validation error shape, matching the `SearchAttacksToolIT` pattern.

[AIML-1353]: https://contrast.atlassian.net/browse/AIML-1353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/vulnerability/VulnerabilityTypeValidator.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/vulnerability/VulnerabilityTypeValidator.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2026 Contrast Security
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.contrast.labs.ai.mcp.contrast.tool.vulnerability;
+
+import com.contrast.labs.ai.mcp.contrast.client.ContrastApiClient;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+import org.springframework.util.StringUtils;
+
+/** Validates vulnerability types against the live organization rule catalogue. */
+@Service
+@RequiredArgsConstructor
+public class VulnerabilityTypeValidator {
+
+  public static final String INVALID_VULN_TYPES_ERROR_FORMAT =
+      "Invalid vulnTypes: %s. Use list_vulnerability_types to discover valid values.";
+
+  private static final String CATALOGUE_UNAVAILABLE_ERROR =
+      "Vulnerability type catalogue is unavailable";
+
+  private final ContrastApiClient contrastApiClient;
+
+  /**
+   * Validates vulnerability types and returns the catalogue's canonical spelling for each value.
+   *
+   * @param vulnTypes vulnerability types supplied by the caller
+   * @return canonical vulnerability types in caller order
+   * @throws Exception when the live catalogue cannot be retrieved
+   * @throws IllegalArgumentException when any supplied value is not in the catalogue
+   */
+  public List<String> validateAndCanonicalize(List<String> vulnTypes) throws Exception {
+    var rules = contrastApiClient.getRules();
+    if (rules == null || rules.getRules() == null) {
+      throw new IllegalStateException(CATALOGUE_UNAVAILABLE_ERROR);
+    }
+
+    var canonicalTypes = new HashMap<String, String>();
+    for (var rule : rules.getRules()) {
+      var name = rule.getName();
+      if (StringUtils.hasText(name)) {
+        var canonical = name.trim();
+        canonicalTypes.put(canonical.toLowerCase(Locale.ROOT), canonical);
+      }
+    }
+    if (canonicalTypes.isEmpty()) {
+      throw new IllegalStateException(CATALOGUE_UNAVAILABLE_ERROR);
+    }
+
+    var canonicalized = new ArrayList<String>();
+    var invalid = new ArrayList<String>();
+    for (var vulnType : vulnTypes) {
+      var normalized = vulnType.trim();
+      var canonical = canonicalTypes.get(normalized.toLowerCase(Locale.ROOT));
+      if (canonical == null) {
+        invalid.add("'" + normalized + "'");
+      } else {
+        canonicalized.add(canonical);
+      }
+    }
+
+    if (!invalid.isEmpty()) {
+      throw new IllegalArgumentException(
+          String.format(INVALID_VULN_TYPES_ERROR_FORMAT, String.join(", ", invalid)));
+    }
+
+    return List.copyOf(canonicalized);
+  }
+}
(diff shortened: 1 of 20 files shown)


contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/vulnerability/VulnerabilityTypeValidator.java
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+    if (canonicalTypes.isEmpty()) {
+      throw new IllegalStateException(CATALOGUE_UNAVAILABLE_ERROR);
+    }
+
+    var canonicalized = new ArrayList<String>();
+    var invalid = new ArrayList<String>();
+    for (var vulnType : vulnTypes) {
+      var normalized = vulnType.trim();
```
Alex-Contrast, line 69: **Low** — a null element in `vulnTypes` produces the opaque error this PR exists to remove.

**Problem.** `vulnType.trim()` throws `NullPointerException` when the list contains a null element. These lists arrive as JSON arrays produced by an LLM agent, and Jackson deserializes a JSON `null` array entry into a `null` list element, so `{"vulnTypes": ["sql-injection", null]}` reaches this loop.

**Impact.** This is worse than a generic NPE because of where it lands. `CursorPaginatedTool` catches `IllegalArgumentException` specifically at `:89` and returns `e.getMessage()` verbatim, which is how the useful error, `Invalid vulnTypes: 'sqli'. Use list_vulnerability_types to discover valid values.`, reaches the agent. An NPE misses that branch and falls through to the generic handler at `:91`:

```java
return CursorToolResponse.error(
    pagination.pageSize(), "An internal error occurred (ref: " + requestId + ")");
```

So a null element returns "An internal error occurred", which is exactly the undiagnosable response AIML-1353 was filed to eliminate.

**Fix.** Treat null as an invalid value rather than guarding past it, so it flows through the mechanism this PR already built:

```java
for (var vulnType : vulnTypes) {
  if (vulnType == null) {
    invalid.add("null");
    continue;
  }
  var normalized = vulnType.trim();
```

The agent then gets `Invalid vulnTypes: null. Use list_vulnerability_types...`, consistent with every other bad value.

Separately and much weaker, the rules loop at `:55-56` has the same shape (`rule.getName()` on a possibly-null element). That one needs TeamServer to emit a null array entry, which I found no evidence it can, and `:50` already guards the realistic null cases. Worth a defensive `rule != null` while you are here, not worth changing anything else.

Alex-Contrast (commented):
Cross-model review of the vulnTypes validation change, triaged against the branch before posting. One finding, low severity but tightly connected to this PR's purpose, so it is worth a look before merge.

Answers to the four things the review specifically checked, all of which held:

- **Fail-closed cost.** Catalogue unavailability throws `IllegalStateException("Vulnerability type catalogue is unavailable")` while bad input throws `IllegalArgumentException` naming the values, so an agent can distinguish "the service is down" from "you typed it wrong". Good.
- **Per-call overhead.** Both tools guard on `params.getVulnTypes() != null && !isEmpty()` before calling the validator, so searches that pass no `vulnTypes` add zero round trips. When `vulnTypes` is present, `getRules()` is fetched per call with no caching. Fine at current usage, worth remembering if filtered searches become common.
- **Verbatim message delivery.** Confirmed at `CursorPaginatedTool:89-90`. This is also why the finding above matters, the NPE path misses exactly this branch.
- **Canonical spelling forwarded.** Confirmed, the validator maps a lowercased key back to `rule.getName().trim()` from the live catalogue, so canonical casing goes downstream regardless of what the caller sent.

Produced with model assistance and verified against the branch by hand.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/vulnerability/VulnerabilityTypeValidator.java
ChrisEdwards: Fixed in ca1568f on the tip branch (AIML-1354-route-coverage-license-403). Null elements now flow through the existing validation mechanism as `invalid.add("null")` (bare, no quotes), so the agent gets `Invalid vulnTypes: null. Use list_vulnerability_types...` instead of an opaque internal error.

Declined the secondary `rule != null` suggestion at the rules loop — no evidence TeamServer emits null entries, and the repo convention avoids defensive guards for undemonstrated scenarios.

Regression test added: `validateAndCanonicalize_should_throw_illegal_argument_when_values_contain_null`.

ChrisEdwards:
I addressed your issues @Alex-Contrast . Please re-review

Alex-Contrast (approved):

PR #226, opened 2026-08-11
AIML-1362 Add /release orchestration skill

ChrisEdwards:
**Jira:** [AIML-1362](https://contrast.atlassian.net/browse/AIML-1362)

## Summary
Adds a `/release` skill that orchestrates the full mcp-contrast release end to end, composing the existing `/update-changelog` and `/test-mcp-server` skills around the Gradle Release workflow.

- One command drives the whole release, with human gates at changelog approval, version choice, changelog PR merge, go/no-go before dispatch, the tracker sweep, and the Slack announcement.
- The flow is changelog-first. The audited draft is the evidence for the version decision, so the release ticket and branch are created already carrying the approved version.
- `/test-mcp-server` now inherits the launching session's model instead of hardcoding Opus and Sonnet.

## Why This Change Exists
The release process lived in RELEASING.md as manual steps plus two disconnected skills. Nothing owned the sequence, the preflight judgment, the version stamp, workflow monitoring, post-release verification, or the Jira and bead sweep, so each release depended on someone remembering all of it. Recovery knowledge, such as the workflow being safe to re-dispatch because it reuses an existing tag at HEAD, was tribal.

An initial ticket-first design had a flaw. The version was chosen from raw PR titles before the changelog audit, so a breaking change discovered during the audit could invalidate a branch and ticket already named with the wrong version. A design-review session settled 23 decisions and inverted the order.

## Approach
Changelog-first sequencing. `/update-changelog` gains a **deferred-landing mode** that drafts, audits, and gets approval without committing, because the landing branch does not exist yet. Its suggested version feeds a three-option gate (major, minor, patch, each with the computed number). Only after the human picks does the skill create the release ticket and the `AIML-<id>-release-X.Y.Z` branch, then land the changelog and stamp commits there.

The semver definition now lives in `/update-changelog` where the suggestion is computed. Majors are removed or renamed tools and hosting-requirement changes (new Java version, Spring Boot major, changed run instructions). `contrast-mcp-core` API changes are explicitly not breaking, its only consumer is internal. Small tool contract changes stay minor because AI consumers adapt.

Automation stops where human judgment or irreversibility begins. The smoke test is advisory, workflow failures always wait for a human (partial publishes to three registries need eyes before re-dispatch), post-verify failures get a suggested remediation path but no automated remediation, and the Slack post to `#_aiml-team` sends only on approval.

### New release skill
- `.claude/skills/release/SKILL.md` (new) — ten phases: preflight (red CI stops unless explicitly overridden, all-internal ranges ask about dependency-bump-only releases), deferred-landing changelog draft, version gate, ticket/branch/landing, smoke test, go/no-go and dispatch with explicit `release_version` and background `gh run watch`, five-asset and attestation and DockerHub and Artifactory verification, GitHub release-notes overwrite from the stamped changelog, pick-list tracker sweep with batch bead confirmation, gated Slack announcement. Includes re-entry guidance for mid-flight failures.

### update-changelog integration
- `.claude/skills/update-changelog/SKILL.md` — deferred-landing mode, the semver definition, and a house-style rule putting hosting-requirement changes under Breaking Changes so that section drives the bump.

### test-mcp-server model inheritance
- `.claude/skills/test-mcp-server/SKILL.md`, `run.sh` — with no `--model` flag the nested instance and its testers inherit the launching session's model, `--model` overrides both. Mode selection is asked via AskUserQuestion, never defaulted. The jar-discovery `ls | grep` was replaced with a glob loop to satisfy the shellcheck pre-commit hook (SC2010).

### Cross-references
- `RELEASING.md` — new Orchestrated Release section, manual steps stay authoritative.
- `CLAUDE.md` — agent-skills entry for `/release`.

## Testing
Docs and skill instructions plus one shell script, no Java changes, so the unit and coverage gates do not apply. `make check` passes. `run.sh` passes shellcheck and `bash -n`. The skill flow itself is validated by design review (23 settled decisions) and two advisor audits that mapped each decision to the files, its first real exercise will be the next release.

- The skill commits the Slack channel name `#_aiml-team` to a public repo. Precedent exists (SECURITY.md publicly names `#contrast-labs`).
- The deferred-landing contract couples `/release` to `/update-changelog`. The mode is documented in both skills so the dependency is visible to future editors.

[AIML-1362]: https://contrast.atlassian.net/browse/AIML-1362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -1,79 +1,96 @@
 ---
 name: release
-description: Orchestrate a full release of mcp-contrast end to end. Preflight checks, release ticket and branch creation, changelog update and version stamp, smoke test, dispatch and monitoring of the Gradle Release workflow, post-release artifact verification, and the Jira/bead sweep. Run when a human asks to cut, make, or ship a release. Human gates at version choice, changelog approval, and dispatch. Optional argument is an explicit version, e.g. /release 2.5.0.
+description: Orchestrate a full release of mcp-contrast end to end, changelog-first. Preflight, deferred-landing changelog draft, version gate, release ticket and branch, smoke test, Gradle Release dispatch and monitoring, artifact verification, release-notes sync, tracker sweep, and Slack announcement. Run when a human asks to cut, make, or ship a release. Human gates at changelog approval, version choice, changelog PR merge, go/no-go, tracker sweep, and the Slack post.
 ---
 
 # Release orchestration
 
 Runs the whole release documented in RELEASING.md, composing `/update-changelog` and
-`/test-mcp-server`. The skill owns the sequence, the preflight judgment, the version
-stamp, workflow dispatch and monitoring, post-release verification, and the tracker
-sweep. Publishing to Artifactory and DockerHub is irreversible, so three human gates
-are mandatory: version choice, changelog approval, and go/no-go before dispatch.
+`/test-mcp-server`. The flow is changelog-first: the audited changelog draft is the
+evidence for the version decision, and the release ticket and branch are created only
+after the human approves the version, so they are born with the right number.
+Publishing to Artifactory and DockerHub is irreversible, so the human gates below are
+mandatory, never skipped.
 
 ## Usage
 
-- `/release` runs the full flow, recommending the next version.
-- `/release X.Y.Z` uses the given version (no leading `v`).
+`/release` takes no arguments. The version always comes from the suggest-and-approve
+gate in Phase 3.
 
 ## Re-entry
 
 Releases fail mid-flight more than anywhere else. On re-entry, detect completed work
-and skip it: if the stamped changelog is already on `origin/main`, go straight to
-Phase 4 or 5; if the release workflow already ran and only verification is missing,
-go to Phase 6. The workflow itself is safe to re-dispatch because it reuses an
-existing `vX.Y.Z` tag at HEAD.
+from the repo and skip it: stamped changelog already on `origin/main` → go to Phase 5
+or 6; release workflow already succeeded → go to Phase 7. The workflow itself is safe
+to re-dispatch because it reuses an existing `vX.Y.Z` tag at HEAD.
 
 ## Phase 1 — Preflight (read-only)
 
 1. `git fetch origin --tags -q`. Confirm the working tree is clean and HEAD matches
    `origin/main`. If not, stop and tell the human.
 2. Confirm CI is green: `gh run list --branch main --workflow build.yml --limit 1
-   --json conclusion,headSha` and check the run covers the current `origin/main` HEAD.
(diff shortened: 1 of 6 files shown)


Alex-Contrast:
```
pr-tools dependency will stall a run. .claude/skills/release/SKILL.md:77 calls /pr-tools:create-pr, and /update-changelog
Step 6 does the same. If that plugin is not installed in the session running the release, Phase 4 dies right after the
branch and commits are created, which is the worst place to stop because the ticket and branch already exist. CLAUDE.md
documents the install commands, so the fix is either running them or having Phase 1 preflight check for the skill and fail
early.

AskUserQuestion at the version gate. .claude/skills/release/SKILL.md:58 and the reworded step 1 in
test-mcp-server/SKILL.md both depend on that tool. Worth confirming it is available wherever this runs, because if it is
blocked the call fails and the model falls back to asking in prose. Not fatal, but it lands on your most important gate.
Plain prose questions would be more robust.

Phase 9 undercounts shipped tickets. Line 143 derives ticket ids from merge commit subjects. In the range since v2.4.0
there is one merge commit, naming AIML-1354, but four tickets actually shipped, 1352 through 1355. The other three landed
without merge commits of their own. Scanning every commit subject for AIML-\d+ instead of only merge subjects fixes it,
otherwise the sweep silently leaves tickets open.

Deferred-landing mode has a dangling report step. /update-changelog Step 7 asks for a PR URL or commit hash, but in
deferred mode nothing has been committed. The mode's skip list covers the Step 1 branch question and all of Step 6, and it
should cover Step 7 too.
```

Alex-Contrast (approved):

ChrisEdwards:
Addressed in 9674b3a and a8f021c:

**#1 (fix)** Added a pr-tools preflight check to Phase 1 (new step 2). The release now fails early if the plugin is missing, before any branch or ticket is created.

**#2 (decline)** `AskUserQuestion` is a built-in Claude Code tool. If a session's permission config blocks it, the model degrades to a prose question naturally. The structured pick-list with version previews is better UX at the most important gate, so keeping it.

**#3 (fix)** Rewrote Phase 9 step 1. The original merge-commit-subject approach missed tickets that landed via stacked-branch merges (which collapse multiple PRs into one commit). The new approach unions two sources: PR titles (keyed by merge SHA appearing in the tag range, no `--base` filter since stacked PRs may not target main) and all commit subjects in the range (catches direct commits and collapsed stack merges). Your suggested commit-subject-only fix would still miss tickets in a collapsed stack merge where only the last PR's key appears in the commit subject.

**#4 (fix)** Added Step 7 to the deferred-landing skip list in update-changelog.

PR #228, opened 2026-08-13
AIML-1379 Add ratcheting ArchUnit architecture tests

ChrisEdwards:
**Jira:** [AIML-1379](https://contrast.atlassian.net/browse/AIML-1379)

## What the rules enforce (plain English)

**Layer enforcement (dependency direction)**

1. **Result models must not depend on tools** — Classes in `result/` cannot import anything from `tool/`.
2. **SDK data models must not depend on tools** — Classes in `sdkextension/data/` cannot import from `tool/`.
3. **Hints must not depend on tools** — The `hints/` package provides guidance to tools but never calls back into them.

**SDK containment**

4. **Only client and sdkextension may touch the Contrast SDK** — No other package can import from `com.contrastsecurity..`. Everything else codes against our own abstractions so the published core stays decoupled from SDK changes.

**Cycle prevention**

5. **Top-level packages must be cycle-free** — `tool`, `client`, `result`, `sdkextension`, and `hints` must form a one-way dependency graph with no circular references.

**Tool domain isolation**

6. **Tool domains must not cross-import** — A tool in `tool/vulnerability/` cannot depend on code in `tool/application/` or any other sibling domain. Shared code belongs in `tool/base/` or `tool/validation/`.

**Structural conventions**

7. **Params classes belong in a params sub-package** — Any `*Params` class inside a tool domain must live in a `params/` sub-package.
8. **Client classes belong in the client package** — Any `*Client` class must live in `client/`.

**Size and complexity guardrails**

9. **No package may have more than 20 top-level classes** — Split bloated packages into sub-packages.
10. **No god objects** — A class cannot have more than 28 non-accessor methods.
11. **No excessive fan-out** — A class cannot depend on more than 30 other project classes.

Rules 1–8 use a freeze-and-ratchet pattern: existing violations are grandfathered in a committed store, but any new violation fails immediately. Rules 9–11 use hard numeric thresholds. The size guardrails start high because they reflect today's codebase. We ratchet them down as we fix violations, then lock them at a reasonable ceiling and hold.

---

## Summary

Adds ArchUnit 1.4.1 to `contrast-mcp-core` with 11 architectural rules that run as plain JUnit tests, wired as a ratchet. Grandfathered violations live in a committed `FreezingArchRule` store, so every rule passes today, any new violation fails immediately, and the store can only shrink as violations get fixed.

- Architecture rules now enforce layering, cycles, tool domain isolation, naming conventions, and size guardrails that no other tool in the pipeline catches
- A new SDK containment rule limits direct `com.contrastsecurity` use to `client` and `sdkextension`, with 151 grandfathered dependency lines as the burn-down list
- Fixing a grandfathered violation needs no test edits, just a store-shrinking run

## Why This Change Exists

Nothing in the pipeline guarded package structure. Result models had grown dependencies back onto tool classes, tool domains had started cross-importing, and SDK types leak through most layers. We wanted rules that block new structural drift immediately while letting the existing debt burn down incrementally, the same model as the repo's coverage and mutation floors.

## Approach

- Dependency and convention rules are wrapped in `freeze(...)`. The store records exact violations (per dependency, not per class), so a previously exempted class cannot quietly add a second violation. Hand-rolled exclusion lists cannot give that precision.
- The store is read-only on every normal run, so CI and PIT can never mutate it. After fixing a violation, run `./gradlew :contrast-mcp-core:test -ParchStoreUpdate` and commit the shrunken store. Store entries may only be removed, never added, without explicit approval, the same policy as checkstyle suppressions (now stated in CLAUDE.md).
- The three size guardrails are not frozen because their failure messages embed live counts. Their thresholds ratchet manually like coverage floors, lowered as improvements land, never raised.

### Rules (`ArchitectureTest.java`)

- Layer enforcement (3 rules), `result`, `sdkextension.data`, and `hints` must not depend on `tool`
- SDK containment (1 rule), only `client` and `sdkextension` may depend on the Contrast SDK directly
- Cycle detection (1 rule), top-level packages must form a DAG
- Tool domain isolation (1 rule), the seven tool domains must not cross-import, expressed as one condition so the store carries the exceptions
- Structural conventions (2 rules), Params placement and Client naming
- Size guardrails (3 rules, unfrozen), max 20 classes per package, 28 non-accessor methods per class, 30 project-class fan-out

### Ratchet plumbing

- `contrast-mcp-core/src/test/resources/archunit.properties`, store location and read-only defaults, plus the documented procedure for adding a new frozen rule
- `contrast-mcp-core/src/test/resources/archunit-store/`, the committed grandfathered violations, 7 result-to-tool lines and the matching package cycle (mcp-r7xbx), 11 coverage-to-application lines (mcp-ax84i), 151 SDK containment lines
- `build.gradle`, the opt-in `-ParchStoreUpdate` flag on test tasks
- `CLAUDE.md`, the ratchet policy, and the store added to the existing prohibition on relaxing checks

## Testing

- `make check-test` passes (1159 tests, all coverage and mutation floors met)
- Enforcement verified with a negative test, removing one line from the store made exactly the owning rule fail, restoring it went green
- A read-only run (no `-ParchStoreUpdate`) passes and leaves the store untouched
- Gap, the shrink direction (`-ParchStoreUpdate` removing fixed entries) is stock ArchUnit behavior and gets its first real exercise when mcp-r7xbx lands

- mcp-r7xbx, move `FilterHelper.formatTimestamp` out of `tool.base`, clears the result-to-tool violations and the one package cycle
- mcp-ax84i, move `ApplicationLicenseDiscriminator` to `tool.base`, clears the cross-domain violations
- mcp-fu1nw, burn down the 151-line SDK containment list, the largest and slowest item
- `stored.rules` carries a `Properties.store` timestamp, so burn-down PRs will include a one-line timestamp diff, inherent to ArchUnit's store format

[AIML-1379]: https://contrast.atlassian.net/browse/AIML-1379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alex-Contrast (commented):
Review of AIML-1379. No inline comments — nothing here needs a thread. One minor note below.

The ratchet design holds up. Freezing per dependency rather than per class is the right precision, keeping the store read-only by default is what stops CI and PIT absorbing new violations, and leaving the three size guardrails unfrozen because their messages embed live counts is a real distinction rather than an inconsistency. Worth noting that four of the seven store files start at zero bytes, so four rules are already clean — a stronger starting position than the PR body claims for itself.

**Deliberately not findings**, recorded so they do not get raised again:

- Whether a new cross-import between two tool domains could be absorbed by a store entry belonging to a different domain pair, given rule 6 collapses seven domains into one condition. `FreezingArchRule` matches violation strings exactly, including source class, target class, signature and line number, so an unmatched pair fails immediately. This was the thing I most wanted to be wrong about and it is fine.
- Threshold headroom on rules 9–11. Current maxima sit below each threshold with room to spare.
- Any risk of `-ParchStoreUpdate` being set on a normal or CI run. `allowStoreUpdate` defaults to false in `archunit.properties` and is only flipped by the explicit flag.

**Open question, genuinely a question rather than a change request:** the shrink direction gets its first real exercise in #235 and #237. Both land removal-only store diffs, which is the good case. Is there anything that would catch a shrink that *silently no-ops* — where a fixed violation is removed from the code but the store entry stays, so the store slowly accumulates dead entries that nothing re-fires against? Nothing in this PR needs to solve that; I ask because the store is the burn-down ledger and a stale ledger reads the same as a clean one.

<details>

- `contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/ArchitectureTest.java:81` — `data_models_should_not_depend_on_tools` (line 81), `hints_should_not_depend_on_tools` (line 91) and `client_classes_should_live_in_client_package` (line 204) omit `.as(...)` while their five siblings supply it, so ArchUnit generates their `stored.rules` keys from the package pattern. You can see the split in the committed store: five clean keys, three that begin `no classes that reside in a package '…'`. Beyond the noisy failure output, the store key is the rule's identity, so for those three a package rename silently changes the key, orphans the store entry and un-freezes its grandfathered violations. Three added lines makes the ratchet durable against renames.
</details>

_Produced with model assistance. One further advisory the reviewer raised — that the fan-out rule may count self-references — was dropped rather than posted, because confirming it requires knowing whether ArchUnit emits self-dependencies from `getDirectDependenciesFromSelf()` and I could not verify that here. It cannot change a pass into a fail at the current headroom either way._

ChrisEdwards:
**Re: missing `.as(...)` on three ArchUnit rules** — fixed in 8fd0586 on #237. All three rules (`data_models_should_not_depend_on_tools`, `hints_should_not_depend_on_tools`, `client_classes_should_live_in_client_package`) now have `.as("...")` descriptions, and the freeze-store keys have been migrated via `-ParchStoreUpdate`.

Also added the `.as()` requirement as a convention in the archunit skill so future rules get it by default.

**Re: stale store entries** — good question. Nothing currently detects a dead store entry (violation fixed in code but entry left behind). The ratchet reads the same either way. Worth addressing but not blocking this stack.

Alex-Contrast (approved):

PR #237, opened 2026-08-14
AIML-1394 Move ApplicationLicenseDiscriminator to tool.base

ChrisEdwards:
**Jira:** [AIML-1394](https://contrast.atlassian.net/browse/AIML-1394)

## Summary
Moved `ApplicationLicenseDiscriminator` from `tool.application` to `tool.base` to eliminate a cross-domain dependency that `tool.coverage` had on `tool.application`. Introduced `ContrastAccessDeniedException` in the `client` package to replace the SDK `UnauthorizedException` parameter, fixing the SDK containment violation at the same time. Added an `archunit` skill and two guardrail hooks that keep the agent from manually editing freeze-store files or attempting fixes without loading the skill first.

- Cross-domain tool violation eliminated (freeze store emptied)
- SDK containment violation fixed via proper abstraction rather than store entry
- One fewer grandfathered violation in the ArchUnit freeze store
- New `archunit` skill codifies the fix-then-shrink workflow and SDK containment patterns
- Two guardrail hooks prevent the agent from taking wrong turns on ArchUnit violations

## Why This Change Exists
- ArchUnit tests (AIML-1379) flagged that `tool.coverage.GetRouteCoverageTool` depended on `tool.application.ApplicationLicenseDiscriminator`, violating tool domain isolation
- The class serves as shared infrastructure (classifying HTTP 403 responses), so `tool.base` is the correct home
- Moving the class also surfaced an SDK containment violation, since the discriminator accepted `UnauthorizedException` (a Contrast SDK type) in its public API
- During the fix work, the agent repeatedly tried to edit freeze-store files manually and attempted fixes without loading the right guidance, wasting turns and producing wrong fixes. The hooks and skill prevent that from happening again.

### SDK boundary
- `ContrastAccessDeniedException` (new) — wraps the SDK `UnauthorizedException` fields so tool-layer code handles access denials without importing the SDK
- `ApplicationLicenseDiscriminator` — moved to `tool.base`, parameter changed from `UnauthorizedException` to `ContrastAccessDeniedException`
- `GetRouteCoverageTool` — converts at the catch site with `ContrastAccessDeniedException.from()`

### ArchUnit store
- `b141356a` (tool domain isolation) — emptied, all 12 entries resolved
- `99efc868` (SDK containment) — one stale `ApplicationLicenseDiscriminator` entry removed
- `ArchitectureTest` Javadoc — removed the now-fixed grandfathered violation mention

### `archunit` skill (`.claude/skills/archunit/SKILL.md`)
Agent-invokable skill that codifies what we learned fixing violations on this branch:
- **Fix-then-shrink workflow** — fix the code, run the store update task, verify with `make check-test`, commit the shrunken store alongside the code changes
- **SDK containment patterns** — wrap SDK types in domain abstractions (`YourType.from(SdkType)`), same pattern for SDK exceptions
- **Layering fix patterns** — `result` depends on `tool` → extract to `util`; tool cross-imports → move to `tool.base`
- **Anti-patterns** — manual store edits, weakening types to dodge a dependency, running `-ParchStoreUpdate` hoping it absorbs new violations

### Guardrail hooks (`.claude/hooks/`, `.claude/settings.json`)
Two hooks keep the agent on the right path when ArchUnit violations appear:

**`archunit-store-guard.sh`** (PreToolUse, matches `Read|Write|Edit|Bash`)
- Blocks Write and Edit to any freeze-store path. Only `./gradlew`, `make`, and `git` commands that touch the store are allowed through.
- Read access shows a warning reminding the agent not to modify the files.

**`archunit-reminder.sh`** (PostToolUse, matches `Bash`)
- Fires when Bash output contains `ArchitectureTest.*FAILED`
- Tells the agent to load the `archunit` skill before attempting any fix

Both hooks include a 13-case smoke test (`test-archunit-hooks.sh`) that verifies deny/allow/warn decisions across Read, Write, Edit, and Bash tool inputs.

### Import updates
- `GetRouteCoverageToolTest`, `GetRouteCoverageLocalParityTest`, `GetRouteCoverageToolIT` — updated imports to new package paths

## Testing
- All 1196 unit tests pass
- Coverage floors met (core 97.9% line / 91.3% branch, stdio-app 97.8% / 96.2%)
- Mutation test strength floor met
- ArchUnit store update verified: `tool_domains_should_not_cross_import` store is now empty
- Existing `ApplicationLicenseDiscriminatorTest` and `GetRouteCoverageToolTest` discrimination tests exercise the new `ContrastAccessDeniedException` parameter path
- Hook smoke tests pass all 13 cases (deny mutations, allow Gradle/make/git, allow reads with warning, ignore non-store paths, fire reminder on ArchUnit failure, stay silent on unrelated output)

[AIML-1394]: https://contrast.atlassian.net/browse/AIML-1394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.claude/hooks/archunit-store-guard.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+if [ "$tool_name" = "Bash" ]; then
+  command=$(echo "$input" | jq -r '.tool_input.command // ""')
+  case "$command" in
+    *archunit-store/*)
+      # Legitimate paths: the Gradle store-update task, make targets that
+      # wrap it, and git operations (diff, add, commit of the shrunken store)
+      case "$command" in
+        ./gradlew*|make*|git\ *) exit 0 ;;
```
Alex-Contrast: **Medium** — any command chained behind `git`, `make`, or `./gradlew` reaches the store unguarded.

**Problem:** this `case` matches a prefix against the *entire* command string, so it allows any chain whose first word is allow-listed, regardless of what follows.

**Impact:** confirmed by running the hook from this branch rather than by reading it. The control is the load-bearing line — the identical redirect denies on its own and is allowed once four characters of `git ` precede it:

```
control (expect deny):
echo pwned > …/archunit-store/b141356a-…                    -> deny

chained behind an allow-listed prefix:
git status; echo pwned > …/archunit-store/b141356a-…        -> ALLOWED (hook exited 0, emitted nothing)
make test && tee …/archunit-store/b141356a-…                -> ALLOWED (hook exited 0, emitted nothing)
./gradlew help; sed -i '' 's/x/y/' …/archunit-store/b141…   -> ALLOWED (hook exited 0, emitted nothing)
```

This matters more than a normal robustness gap because of why the hook exists. The PR body says the agent "repeatedly tried to edit freeze-store files manually" — an agent that reaches for a store edit and gets denied is very likely to retry with a chained command, which is the exact shape that passes.

**Fix:** split the command on `&&`, `||`, `;` and `|` and apply the allow-list to each segment, rather than matching the whole string. Worth a suite case per bypass shape too — the current 13 cases are sound, they just have no case for a chain.

.claude/hooks/archunit-store-guard.sh
```diff
@@ -0,0 +1,43 @@
...
+
+deny() {
+  jq -n --arg reason "$reason" '{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny","permissionDecisionReason":$reason}}'
+  exit 0
+}
+
+if [ "$tool_name" = "Bash" ]; then
+  command=$(echo "$input" | jq -r '.tool_input.command // ""')
+  case "$command" in
+    *archunit-store/*)
```
Alex-Contrast, line 19: **Medium** — a command targeting the store *directory* is not seen by the guard at all.

**Problem:** `*archunit-store/*` requires at least one character after the slash. A command that names the directory and stops there never matches this outer `case`, so it falls straight through to the unconditional `exit 0` at the end of the Bash branch — it is not denied, and it is not even evaluated against the allow-list.

**Impact:** confirmed by running the hook:

```
rm -rf contrast-mcp-core/src/test/resources/archunit-store        -> ALLOWED (hook exited 0, emitted nothing)
cp /tmp/evil contrast-mcp-core/src/test/resources/archunit-store  -> ALLOWED (hook exited 0, emitted nothing)
```

`rm -rf` on the store directory is the worst outcome this guard exists to prevent, and it is the case that passes most cleanly.

**Fix:** widen to `*archunit-store*` here. The same pattern appears again at line 34 for the `file_path` branch; that one matters much less, since Write and Edit target files rather than directories, but it is worth changing together for consistency. Note this fix does not address the chaining bypass in the other comment — they are independent.

Alex-Contrast (commented):
Review of AIML-1394. Two inline comments, both on `.claude/hooks/archunit-store-guard.sh`, both independent of each other and both confirmed by running the hook rather than by reading it.

The class move and the `ContrastAccessDeniedException` abstraction both check out and have no comments. Introducing a real domain type rather than adding a freeze-store entry is the right call and matches what #228's policy asks for.

**Deliberately not findings**, so they do not get re-litigated:

- `ContrastAccessDeniedException.from()` dropping a field the discriminator needs. Checked and cleared. `ApplicationLicenseDiscriminator.discriminate()` does not inspect exception fields at all, so the wrapper cannot lose one, and `GetRouteCoverageToolTest`'s new `ArgumentCaptor` assertion pins status code and cause anyway.
- The 13-case hook suite being deny-only. Checked and cleared — stubbing the guard to deny unconditionally fails 5 of 9 guard cases, so the allow cases are real. The suite is sound; it simply has no case for either bypass shape, which is what the two comments ask for.
- The import-only test updates. No assertion was weakened.

**Open question, genuinely a question:** is the Bash arm of this guard meant to be a security boundary or a speed bump? If it is a speed bump for a cooperating agent that occasionally reaches for the wrong tool, the current shape is arguably adequate and both comments drop to nits. If it is meant to hold against an agent that retries after a denial, it needs the segment-splitting fix. The PR body's account of the agent "repeatedly" trying store edits is what makes me think it is closer to the second.

_Produced with model assistance. The bypass evidence above is actual output from this branch's hook, not a description of what it would do._

.claude/hooks/archunit-store-guard.sh
ChrisEdwards: Fixed in 8fd0586. The guard now splits the command on `;`, `&&`, `||`, and `|` and checks each segment independently against the allow-list. Added 4 smoke test cases covering each chaining shape — all deny as expected.

.claude/hooks/archunit-store-guard.sh
ChrisEdwards: Keeping the original trailing-slash glob (`*/`) rather than widening it. The broader match causes false positives on commands that merely mention the store by name (e.g. PR comment replies). The directory-level risk is low and git restores the directory trivially. The command-chaining fix in the sibling thread is the meaningful hardening.

Alex-Contrast (approved):

PR #239, opened 2026-08-18
AIML-1227 Add open-crap4j quality gate

ChrisEdwards:
**Jira:** [AIML-1227](https://contrast.atlassian.net/browse/AIML-1227)

## Summary

Integrates the open-crap4j Gradle plugin into the build, giving every module CRAP (Change Risk Anti-Patterns) scoring on top of existing JaCoCo coverage data. Two known violations are baselined.

- CRAP scoring now available per-module via `./gradlew crapReport` (advisory) and `./gradlew crapCheck` (enforcing)
- Two core methods baselined (`RecommendationMarkdownRenderer.registerKnownTags` at 38.50, `SearchAppVulnerabilitiesTool.doExecute` at 18.52)
- stdio-app module is clean with no violations

## Why This Change Exists

The build measures line/branch coverage but has no signal on methods that are both complex and undertested. An ad-hoc run found 15 methods above CRAP 8.0 out of 387 scored. Standing CRAP reports surface that signal without a new runtime dependency, since JaCoCo XML already carries per-method complexity and coverage counters.

## Approach

Uses the open-crap4j Gradle plugin (`com.architester.crap4j` version `1.0.0`) from the Gradle Plugin Portal. The plugin owns the CRAP calculation, baseline workflow, enforcing/advisory tasks, JSON output, and JUnit sidecars, so there is nothing to reimplement. Changed-file analysis stays in the open-crap4j CLI rather than the Gradle build.

### Build integration
- `build.gradle` — declares the plugin and applies it to all subprojects after JaCoCo config

### Baselines
- `contrast-mcp-core/crap4j-baseline.json` — two baselined methods
- `contrast-mcp-stdio-app/crap4j-baseline.json` — empty baseline (clean module)

### Documentation
- `CLAUDE.md` — documents CRAP gate usage
- `docs/research/open-crap4j-local-integration.md` — research notes on plugin contract, lifecycle risks, and dogfooding results

## Testing

Validated by running `./gradlew crapCheck` end to end. The plugin analyzed 383 core methods and 84 stdio methods. After baselining, `crapCheck` passes, `crapBaselineTighten` confirms baselines are tight, and both JSON and JUnit sidecars are well-formed. CLI changed-file checks passed against the real stack diff.

No new automated tests in this repo. The plugin itself carries its own functional test suite.

## Risks / Rollout / Follow-ups

- `crapCheck` is not wired into `make check`, `make check-test`, or CI. That is a deliberate follow-up after the plugin is published.
[...]

PR #241, opened 2026-08-18
AIML-1407 Standardize verification on Gradle check and verify lifecycles

ChrisEdwards:
**Jira:** [AIML-1407](https://contrast.atlassian.net/browse/AIML-1407)

## Summary

Make targets, CI workflows, and the pre-push hook each enumerated Gradle task names by hand, so every new verification had to be added to every invoker separately. The crap4j gate (AIML-1227) proved the failure mode: `crapCheck` was registered but no gate ran it, pitest ran on PRs but not releases, and the release workflow invoked `integrationTest` with no credentials so every IT silently skipped. This PR standardizes on two Gradle lifecycle tasks as the single source of truth, eliminating enumeration drift and making integration tests actually gate releases for the first time.

## Why This Change Exists

The crap4j work (AIML-1227) exposed a structural problem: adding a new verification step to Gradle meant hunting down every place that enumerated task names and updating each one independently. The invokers had already drifted from each other. PIT mutation testing ran on PR CI but not release CI. The release workflow called `integrationTest` but never set credentials, so it went green having run zero tests. The repo's own testing standards forbid silent skips, and the build itself was the violator.

The pre-push hook only checked changed-file coverage, not the full gate, so a developer who skipped `make check-test` before pushing could land code that failed CI. The Makefile had grown to seven overlapping targets (`check`, `test`, `coverage`, `test-coverage`, `check-test`, `mutation`, `verify`) that each defined their own version of "verified," and none of them matched what CI actually ran.

## The Big Picture

```mermaid
flowchart LR
    subgraph "Before: enumerated task lists"
        M1["make check-test<br/>(4 sub-makes)"] --> G1["spotlessCheck<br/>checkstyleMain/Test<br/>test<br/>jacocoVerify<br/>pitest"]
        CI1["build.yml"] --> G2["spotlessCheck<br/>checkstyleMain/Test<br/>jacocoVerify<br/>bootJar"]
        CI1 --> G3["pitest (PR only)"]
        REL1["gradle-release.yml"] --> G4["spotlessCheck<br/>check<br/>integrationTest<br/>(no creds)"]
    end

    subgraph "After: lifecycle tasks"
        M2["make check"] --> GC["gradlew check<br/>(all repo-local verification)"]
        M3["make verify"] --> GV["gradlew verify<br/>(check + ITs w/ creds)"]
        CI2["build.yml"] --> GC
        REL2["gradle-release.yml"] --> GV
        HOOK["pre-push"] --> GC
    end
```

Key decisions:

- **PIT inside `check`**, not carved out as a slow-path extra. A full PIT run is 38s locally and cached by Gradle up-to-date checks when core is unchanged. Carving it out recreates the enumeration disease for one task.
- **`make check` is the full gate, not the fast loop.** A make target named `check` that does less than `gradlew check` is the ambiguity this decision removes. The fast inner-loop path gets its own name (`lint`).
- **`verify` fails loudly without credentials.** A gate that can pass on zero integration tests is not a gate. Bare `integrationTest` keeps skip behavior for forks and credential-less checkouts.
- **`verify` in no hook.** Live-API tests are minutes long, network-dependent, and occasionally flaky for reasons unrelated to the change. Hooks that fail spuriously teach people to bypass hooks.
- **Gradle sources `.env.integration-test` itself.** Real environment variables win over file values, so CI secrets and one-off shells override without editing the file. No `source` step needed.

### 1. ADRs and glossary — **Focus**
The decision records capture the full rationale. ADR 0003 covers the check/verify standardization; ADR 0004 covers the integration-test fail-loud policy. CONTEXT.md gains glossary entries for Check, Verify, and Lint so the ubiquitous language is pinned.

- `docs/adr/0003-standardize-on-gradle-check-and-verify-lifecycles.md` (new)
- `docs/adr/0004-integration-tests-fail-loudly-and-gate-releases.md` (new)
- `CONTEXT.md` — new Build verification section

### 2. Gradle lifecycle wiring — **Focus**
The core of the change: `crap4j { attachToCheck = true }` in root `build.gradle`, PIT wired into `check` in `contrast-mcp-core/build.gradle`, and the new root `verify` task plus credential sourcing/guarding in `contrast-mcp-stdio-app/build.gradle`.

- `build.gradle` — crap4j attachToCheck, new `verify` task
- `contrast-mcp-core/build.gradle` — `check.dependsOn pitest`
- `contrast-mcp-stdio-app/build.gradle` — `.env.integration-test` parser, `resolveIntegrationTestCredentials`, `requireIntegrationTestCredentials` task

### 3. CI workflows — **Focus**
PR CI collapses its enumerated task list to `gradlew check` plus `bootJar`. The separate PIT step is removed (folded into `check`). Release CI switches from the drifted list to `gradlew clean verify bootJar` with the five `CONTRAST_*` secrets wired in as environment variables.

- `.github/workflows/build.yml` — simplified to `check` + `bootJar`
- `.github/workflows/gradle-release.yml` — `verify` with credential env vars

### 4. Makefile — **Focus**
Seven overlapping targets collapse to thin lifecycle wrappers. `check` and `verify` each delegate to one Gradle lifecycle, with `spotlessApply` prepended and `coverageSummary` appended. `lint` is the new fast inner loop (format + checkstyle). Deleted: `check-test`, `test`, `test-coverage`, `coverage`, `mutation`.

- `Makefile`

### 5. Pre-push hook — **Skim**
Gains `gradlew check` (up-to-date checking makes it a few seconds when `make check` already ran). `SKIP_PUSH_GATE=1` replaces the old `SKIP_COVERAGE_HOOK` as the umbrella bypass.

- `scripts/git-hooks/pre-push`
- `scripts/git-hooks/README.md`

<details>
<summary>Documentation updates (10 files)</summary>

References to the old targets (`make check-test`, `make test`, `make coverage`, `make mutation`) updated throughout:
- `CLAUDE.md` — build commands, coverage, mutation testing, integration tests, testing requirements
- `REPO-CONTEXT.md` — quick-reference commands
- `CONTRIBUTING.md` — common commands, test coverage section
- `INTEGRATION_TESTS.md` — simplified setup (no `source` step), reordered commands, updated behavior notes
- `RELEASING.md` — validation gate description
- `.claude/commands/improve-harness.md` — `check-test` → `check`
- `.claude/settings.json` — allowlist entries
- `.claude/skills/archunit/SKILL.md` — `check-test` → `check`
- `.claude/skills/bead-workflow/SKILL.md` — `check-test` → `check`
- `.claude/skills/test-mcp-server/SKILL.md` — `check-test` → `check`

</details>

## Testing

Verified locally:
- `make check` passes (1196 tests, includes static analysis, coverage floors, CRAP gate, PIT mutation testing)
- `make verify` passes (1430 tests, including 234 live integration tests against eval org, 112s total)
- `requireIntegrationTestCredentials` fails loudly when credentials are removed (tested by unsetting env vars)
- Gradle `--dry-run` on `check` confirms `crapCheck` and `pitest` are in the task graph
- Pre-push hook ran `gradlew check` successfully on push (visible in push output above)

- **Main-branch CI gains ~3 minutes** from PIT (previously PR-only). Accepted as the cost of consistent gating.
- **Release CI credentials** are initially a maintainer's personal eval-org credentials. AIML-1408 (high priority, blocked on this PR) migrates them to a dedicated service account.
- **Pre-push hook is slower** on first push (~60s cold when core changed, ~6s warm). `SKIP_PUSH_GATE=1` is the emergency bypass.

[AIML-1407]: https://contrast.atlassian.net/browse/AIML-1407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/buildSrc/src/test/java/com/contrast/labs/build/EnvFileParserTest.java
+++ b/buildSrc/src/test/java/com/contrast/labs/build/EnvFileParserTest.java
@@ -1,0 +1,144 @@
+package com.contrast.labs.build;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class EnvFileParserTest {
+
+  @TempDir Path tmp;
+
+  @Test
+  void parseLines_should_return_empty_map_when_no_lines() {
+    assertThat(EnvFileParser.parseLines(List.of())).isEmpty();
+  }
+
+  @Test
+  void parseLines_should_skip_comment_lines() {
+    var result = EnvFileParser.parseLines(List.of("# this is a comment", "KEY=value"));
+    assertThat(result).containsExactly(Map.entry("KEY", "value"));
+  }
+
+  @Test
+  void parseLines_should_skip_blank_lines() {
+    var result = EnvFileParser.parseLines(List.of("", "  ", "KEY=value"));
+    assertThat(result).containsExactly(Map.entry("KEY", "value"));
+  }
+
+  @Test
+  void parseLines_should_strip_export_prefix() {
+    var result = EnvFileParser.parseLines(List.of("export FOO=bar"));
+    assertThat(result).containsExactly(Map.entry("FOO", "bar"));
+  }
+
+  @Test
+  void parseLines_should_strip_double_quotes_from_values() {
+    var result = EnvFileParser.parseLines(List.of("KEY=\"quoted value\""));
+    assertThat(result).containsExactly(Map.entry("KEY", "quoted value"));
+  }
+
+  @Test
+  void parseLines_should_strip_single_quotes_from_values() {
+    var result = EnvFileParser.parseLines(List.of("KEY='quoted value'"));
+    assertThat(result).containsExactly(Map.entry("KEY", "quoted value"));
+  }
+
+  @Test
+  void parseLines_should_not_strip_mismatched_quotes() {
+    var result = EnvFileParser.parseLines(List.of("KEY=\"mismatched'"));
+    assertThat(result).containsExactly(Map.entry("KEY", "\"mismatched'"));
+  }
+
+  @Test
+  void parseLines_should_not_strip_single_char_quoted_value() {
+    var result = EnvFileParser.parseLines(List.of("KEY=\""));
+    assertThat(result).containsExactly(Map.entry("KEY", "\""));
+  }
+
+  @Test
+  void parseLines_should_handle_empty_value() {
+    var result = EnvFileParser.parseLines(List.of("KEY="));
+    assertThat(result).containsExactly(Map.entry("KEY", ""));
+  }
+
+  @Test
+  void parseLines_should_skip_lines_without_separator() {
+    var result = EnvFileParser.parseLines(List.of("NOSEP", "KEY=value"));
+    assertThat(result).containsExactly(Map.entry("KEY", "value"));
+  }
+
+  @Test
+  void parseLines_should_skip_lines_with_leading_equals() {
+    var result = EnvFileParser.parseLines(List.of("=value", "KEY=value"));
+    assertThat(result).containsExactly(Map.entry("KEY", "value"));
+  }
+
+  @Test
+  void parseLines_should_use_first_equals_as_separator() {
+    var result = EnvFileParser.parseLines(List.of("KEY=val=ue"));
+    assertThat(result).containsExactly(Map.entry("KEY", "val=ue"));
+  }
+
+  @Test
+  void parseLines_should_trim_whitespace_around_key_and_value() {
+    var result = EnvFileParser.parseLines(List.of("  KEY  =  value  "));
+    assertThat(result).containsExactly(Map.entry("KEY", "value"));
+  }
+
+  @Test
+  void parseLines_should_handle_export_with_quotes() {
+    var result = EnvFileParser.parseLines(List.of("export KEY=\"hello world\""));
+    assertThat(result).containsExactly(Map.entry("KEY", "hello world"));
+  }
+
+  @Test
+  void parseLines_should_preserve_order() {
+    var result = EnvFileParser.parseLines(List.of("A=1", "B=2", "C=3"));
+    assertThat(result.keySet()).containsExactly("A", "B", "C");
+  }
+
+  @Test
+  void parse_should_return_empty_map_when_file_does_not_exist() {
+    assertThat(EnvFileParser.parse(new File(tmp.toFile(), "missing"))).isEmpty();
+  }
+
+  @Test
+  void parse_should_read_file_contents() throws IOException {
+    Path envFile = tmp.resolve("test.env");
+    Files.writeString(envFile, "export HOST=example.com\nAPI_KEY='secret'\n");
+    var result = EnvFileParser.parse(envFile.toFile());
+    assertThat(result)
+        .containsEntry("HOST", "example.com")
+        .containsEntry("API_KEY", "secret")
+        .hasSize(2);
+  }
+
+  @Test
+  void parseLines_should_handle_template_format() {
+    var result =
+        EnvFileParser.parseLines(
+            List.of(
+                "# Integration Test Environment Variables",
+                "# Copy this file and fill in your credentials",
+                "",
+                "export CONTRAST_HOST_NAME=app.contrastsecurity.com",
+                "export CONTRAST_API_KEY=your-api-key-here",
+                "export CONTRAST_SERVICE_KEY=your-service-key-here",
+                "export CONTRAST_USERNAME=your-username-here",
+                "export CONTRAST_ORG_ID=your-org-id-here",
+                "",
+                "# Note: .env.integration-test is in .gitignore"));
+    assertThat(result)
+        .hasSize(5)
+        .containsEntry("CONTRAST_HOST_NAME", "app.contrastsecurity.com")
+        .containsEntry("CONTRAST_API_KEY", "your-api-key-here")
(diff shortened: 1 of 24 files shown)


buildSrc/src/test/java/com/contrast/labs/build/EnvFileParserTest.java
```diff
@@ -0,0 +1,144 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class EnvFileParserTest {
```
Alex-Contrast, line 14: **Low** — `EnvFileParser.resolve` is the only implementation of the env-wins precedence contract, and it has no test.

**Problem:** All 18 tests in this class target `parseLines` (16) or `parse` (2). None calls `resolve`. `resolve` is the method `contrast-mcp-stdio-app/build.gradle` calls, and it is the sole implementation of the "real environment variables win over file values" rule stated in the PR body, in AIML-1407, and in `resolve`'s own Javadoc.

**Impact:** The documented precedence can be reversed with zero test failures. Inverting it so file values win:

```java
String fileValue = fileValues.get(name);
resolved.put(name, fileValue != null ? fileValue : envValue);
```

`./gradlew -p buildSrc test --rerun-tasks` → `BUILD SUCCESSFUL`, 53/53 green (`EnvFileParserTest` 18/0/0, `JacocoChangedFileCoverageTest` 22/0/0, `ChangedJavaFilesTest` 13/0/0).

This is worth a test rather than a shrug because the release gate depends on the precedence: CI secrets are meant to override a committed `.env.integration-test`, so if it ever inverts, a stale file value beats the repo secret and `verify` runs the integration tests against the wrong org while still going green.

**Fix:** Add tests for `resolve` covering the env-override case, the file-fallback case, and a key absent from both.

scripts/git-hooks/pre-push
```diff
@@ -141,8 +141,9 @@ main() {
   local -a remote_refs=()
   local -a remote_shas=()
 
-  if [[ "${SKIP_COVERAGE_HOOK:-}" == "1" ]]; then
-    printf "SKIP_COVERAGE_HOOK=1 set; skipping changed-file coverage verification.\n" >&2
+  # Umbrella bypass, humans only.
+  if [[ "${SKIP_PUSH_GATE:-}" == "1" ]]; then
```
Alex-Contrast, line 145: **Low** — `SKIP_COVERAGE_HOOK` is removed rather than deprecated, contrary to the PR description.

**Problem:** The description says `SKIP_PUSH_GATE=1` replaces `SKIP_COVERAGE_HOOK` as the umbrella bypass "with a deprecation message for the old name". There is no such message. After this PR, `grep -rn "SKIP_COVERAGE_HOOK"` across the tree returns nothing — the name is gone, not deprecated. On `main` and on this PR's base it was a documented bypass in three places: `scripts/git-hooks/pre-push:144-145`, `scripts/git-hooks/README.md:52`, and `CLAUDE.md:20` and `:97`.

**Impact:** Larger than a missing warning, because the same PR makes the hook much heavier by adding `./gradlew -q check` (~60s cold, by the description's own numbers). Anyone reaching for the bypass they have used before gets the old variable silently ignored and then the full gate they were trying to skip, with nothing telling them which variable to use now. That is the "hooks that fail spuriously teach people to bypass hooks" failure mode this PR argues against elsewhere.

**Fix:** Detect `SKIP_COVERAGE_HOOK`, print that it is retired and that `SKIP_PUSH_GATE=1` replaces it, and **still run the gate**. Deliberately not honouring the old name as a working bypass: `CLAUDE.md:20` is an absolute rule against skipping hooks and now names `SKIP_PUSH_GATE=1` as the thing not to use, so reviving the retired name as a second live bypass would cut against that.

Alex-Contrast (commented):
Cross-model review of the `check`/`verify` standardization. Two Low findings, both inline, both about things the PR states but does not implement. Everything else in the change verified clean.

Checked and holding, recorded so they are not re-litigated:

- `verify` cannot pass on zero integration tests. `requireIntegrationTestCredentials` rejects missing *and* empty-string values for all five `CONTRAST_*` keys before `integrationTest` runs.
- No credential value reaches a log. `EnvFileParser` does no logging, the guard's failure message joins only missing key names, and `gradle-release.yml` passes the secrets as step env vars.
- `crapCheck` on the `check` critical path is safe. #239 resolves `com.architester.crap4j:1.0.0` from the plugin portal, not from a composite build, so CI and the pre-push hook can resolve it. (The composite build described in #239's body is no longer what that branch does — noted separately on #239.)

Decisions taken as settled, not re-proposed: PIT inside `check`; `make check` as the full gate with `lint` as the fast loop; `verify` in no hook; the ~3 minute main-branch CI cost; the seven deleted make targets.

Produced with model assistance and verified against the branch at `eb518e905f3e5e7ab36cf76f0b0995e2201961f9`.

buildSrc/src/test/java/com/contrast/labs/build/EnvFileParserTest.java
ChrisEdwards: Fixed in 934919f on the tip branch (`AIML-1412-simplify-high-crap-methods`). Added three tests for `resolve`: env-override (uses `HOME` to verify real env var wins over file value), file-fallback (custom key absent from env falls back to file), and absent-from-both (returns null). All pass in `./gradlew -p buildSrc test`.

scripts/git-hooks/pre-push
ChrisEdwards: Declined. `SKIP_COVERAGE_HOOK` was never used by anyone besides the developer who introduced it. No backwards compatibility is needed. Updated the PR description to remove the inaccurate "with a deprecation message for the old name" text.

Alex-Contrast (approved):

PR #243, opened 2026-08-18
AIML-1410 Pilot property-based testing with jqwik

ChrisEdwards:
**Jira:** [AIML-1410](https://contrastsecurity.atlassian.net/browse/AIML-1410)

## Summary

Adds property-based testing to `contrast-mcp-core` using jqwik 1.10.1, piloted on the validation specs and pagination params. Seven `*PropertyTest` classes check invariants across generated inputs that example-based tests cannot cover. The property tests immediately found two real bugs in `PaginationParams`, both fixed in this PR.

- Two bugs found and fixed by property testing (see below)
- Property tests integrated into the ordinary `check` lifecycle (JaCoCo, CRAP, PIT)
- Core branch coverage rose to 91.3% (floor 87%)
- ADR 0005 records the jqwik adoption decision and Platform-6 compatibility risk

## Why This Change Exists

The validation and pagination code accepts arbitrary integers, strings, and enums. Example-based tests verify chosen inputs but cannot cover the full input space. Edge cases like integer overflow at pagination boundaries, mixed-case enum parsing, and `else-if` bound checking gaps are tedious to enumerate by hand. Property-based testing generates hundreds of input combinations automatically and checks that stated invariants hold for all of them.

## Approach

jqwik 1.10.1 was chosen because it is the only mature property-testing library for the JUnit Platform. A spike confirmed it runs on this repo's Platform 6.0.3 stack via binary compatibility (jqwik targets Platform 1.14.4, which gets force-upgraded). The maintenance risk (single maintainer, uncertain Platform 6 future) is accepted because the fallback is mechanical conversion to parameterized JUnit tests.

Property tests live in the ordinary test source set so the Gradle `check` lifecycle inherits them without enumerating task names (ADR 0003). The default tries budget is 100 per property (`junit-platform.properties`) to keep PIT runtime manageable.

## Bugs Found and Fixed

Property testing found two bugs in `PaginationParams` that existing example-based tests missed.

**1. Default pageSize ignores maxPageSize.** `PaginationParams.of(page, null, 25)` returned `pageSize=50` instead of 25. When `pageSize` was null, the code used `DEFAULT_PAGE_SIZE` (50) without checking `maxPageSize`. A tool with a stricter API limit (e.g. 25) would send an oversized request. The existing test `testCustomMaxPageSizeWithDefaultPageSize` passed by coincidence because it tested with `maxPageSize=50`, which equals `DEFAULT_PAGE_SIZE`. Fixed by using `Math.min(DEFAULT_PAGE_SIZE, maxPageSize)` for the default, matching `CursorPaginationParams` which already had the correct behavior.

**2. Offset integer overflow.** `(actualPage - 1) * actualSize` could overflow `int` for large page values (e.g. `page=50_000_000, pageSize=100` produces `4,999,999,900` which wraps negative). No existing test covered this because example-based tests used small page numbers. Fixed by capping `page` to `Integer.MAX_VALUE / pageSize` with a notice. Property test now proves `offset >= 0` for all inputs.

### Infrastructure
- `gradle.properties` — pinned `jqwikVersion=1.10.1` (released 2026-05-29, past soak window)
- `contrast-mcp-core/build.gradle` — added `testImplementation` for jqwik
- `junit-platform.properties` — created with `jqwik.tries.default=100`
- `.gitignore` — added `.jqwik-database` (jqwik's local failure-recording database)

### Bug fixes (`PaginationParams.java`)
- Default pageSize now clamped to `Math.min(DEFAULT_PAGE_SIZE, maxPageSize)`
- Page capped to `Integer.MAX_VALUE / pageSize` to prevent offset overflow

### Property tests (`tool/base`)
- `PaginationParamsPropertyTest` — clamping bounds, offset identity, offset non-negative, idempotence, notice-iff-corrected, immutability
- `CursorPaginationParamsPropertyTest` — clamping, blank-to-null normalization, cursor presence totality, default-size invariant

### Property tests (`tool/validation`)
- `IntSpecPropertyTest` — output in [min,max], notice count 0 or 1, default handling, no errors for any input
- `StringListSpecPropertyTest` — subset-of-allowed, case-insensitive normalization, idempotence, uppercase conversion
- `EnumSetSpecPropertyTest` — round-trip identity, case-insensitive parsing, deduplication, unknown tokens produce error+null
- `ToolSortParserPropertyTest` — wire-name round-trip, case-insensitive on both property and direction, malformed always yields default
- `ToolValidationContextPropertyTest` — isValid iff errors empty, append-only ordering, requireUuid accepts UUID.toString, immutability

### Documentation
- `docs/adr/0005-adopt-jqwik-for-property-based-testing.md` — records the adoption decision, Platform-6 pairing risk, and fallback plan
- `CLAUDE.md` — added property-testing conventions section

## Testing

All 79 property tests pass across seven `*PropertyTest` classes. `make check` green with 1267 total tests, 0 failures. Coverage floors, CRAP gate, and PIT mutation testing all pass.

Each `@Property` method states one invariant and jqwik generates 100 random inputs per property. The two bugs were caught by properties that would have been difficult to express as example-based tests: "pageSize is always within [1, maxPageSize] for any combination of inputs" and "offset is always non-negative for any page and pageSize."

---
Generated with Claude Code

[AIML-1410]: https://contrast.atlassian.net/browse/AIML-1410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,6 @@ todos/
 .bv/
 
 .orca/
+
+# jqwik failure-recording database
+.jqwik-database
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -347,6 +347,15 @@ When creating or modifying MCP tools:
 - Manual loop for sort/filter verification (e.g., `for (int i=1; i<list.size(); i++)`) — use `isSortedAccordingTo` / `allSatisfy` for clearer failure messages
 - `_populate_*` or `_return_*` test using `isNotNull()` on the claimed field — asserts shape, not content; use `isNotBlank` / `isNotEmpty` / content-level check
 
+**Property-based testing (jqwik, ADR 0005):**
+- Naming: `XxxPropertyTest` alongside existing `XxxTest`, same package. Never `*IT` / `*IntegrationTest` suffixes.
+- Each `@Property` method tests one invariant that holds for all inputs (e.g. "output is always in [min, max]"). Use `@Provide` methods for custom generators.
+- Default tries budget is 100 (`junit-platform.properties`). Keep it modest because PIT reruns tests per mutant.
+- Use `Assume.that(...)` to narrow generators to documented preconditions (e.g. `min <= max`), not to skip inconvenient failures.
+- A failing property is a bug until proven otherwise. Fix the bug, or narrow the generator to documented bounds with a WHY comment and a filed bead.
+- Property tests complement example-based tests. Do not replace existing `XxxTest` classes.
+- `.jqwik-database` is gitignored (jqwik writes a local failure-recording database).
+
 ### Security Considerations
 
 This codebase handles sensitive vulnerability data. The README contains critical warnings about data privacy when using with AI models. Never expose Contrast credentials or vulnerability data to untrusted AI services.
--- a/contrast-mcp-core/build.gradle
+++ b/contrast-mcp-core/build.gradle
@@ -180,7 +180,7 @@ pitest {
     junit5PluginVersion = pitestJunit5PluginVersion
     targetClasses = ['com.contrast.labs.ai.mcp.contrast.*']
     targetTests = ['com.contrast.labs.ai.mcp.contrast.*']
-    testStrengthThreshold = 92
+    testStrengthThreshold = 93
     outputFormats = ['HTML', 'XML']
     timestampedReports = false
     threads = Runtime.runtime.availableProcessors()
--- a/contrast-mcp-core/build.gradle
+++ b/contrast-mcp-core/build.gradle
@@ -23,6 +23,7 @@ dependencies {
 
     testImplementation 'org.mockito:mockito-core'
     testImplementation 'ch.qos.logback:logback-classic'
+    testImplementation "net.jqwik:jqwik:${jqwikVersion}"
 }
 
 def publicMavenReleaseRepositoryName = 'contrastPublicRelease'
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/PaginatedTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/base/PaginatedTool.java
@@ -66,7 +66,7 @@ public abstract class PaginatedTool<P extends ToolParams, R> extends BaseTool {
    * for tool-specific logic. This method is FINAL to enforce consistent processing.
    *
    * @param page requested page number (1-based), null defaults to 1
-   * @param pageSize requested page size, null defaults to 50
+   * @param pageSize requested page size, null defaults to 50 unless getMaxPageSize() is narrower
    * @param paramsSupplier lazy supplier of tool-specific parameters
    * @return paginated response with items or errors
    */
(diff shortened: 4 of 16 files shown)


Alex-Contrast (commented):
Cross-model review of the jqwik pilot. No functional findings. One note below.

The two production bug fixes in `PaginationParams` were the focus, since the properties are the only thing pinning them. Both hold:

- The offset-overflow fix is correct at its boundaries. `maxPageSize` is clamped to at least 1 and `actualSize` to `[1, maxPageSize]` before `maxPage` is computed, so the division cannot be by zero; `offset >= 0` holds for every `int` pair, not just generated ones.
- Both properties that pin the fixes can actually fail. `pageSize_should_respect_custom_max` reaches `pageSize = null` with `maxPageSize < 50`, and `offset_should_never_overflow` reaches `page` values up to `Integer.MAX_VALUE`. Neither is vacuous.
- `junit-platform.properties` is on the test runtime classpath at `contrast-mcp-core/src/test/resources/`, so `jqwik.tries.default=100` is genuinely in force rather than jqwik's own default.

On the `IntSpec` inverted-range case AIML-1410 lists as suspected: `IntSpecPropertyTest` filters it out with `Assume.that(min <= max)` and `IntSpec.java` is untouched, which matches the ticket putting that one in follow-up scope. Noted rather than raised.

Taken as settled and not re-proposed: jqwik on Platform 6.0.3 as a binary-compatibility pairing with a documented fallback (ADR 0005); property tests in the ordinary test source set; 100 tries; clamping the default `pageSize` to match `CursorPaginationParams`; capping `page` rather than widening `offset` to `long`.

Produced with model assistance and verified against the branch at `5803201e5b7af5a42485b1abe42841a9a4384b70`.

<details>

- `contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/validation/ToolSortParserPropertyTest.java:100` — the `if (ctx.isValid() && result != null)` guard is dead code: `validSortString` composes `validProperty()` with `ASC`/`DESC`, so every generated input is valid and the guard is never false. It would suppress a future regression rather than report one, but nothing is actually at risk, because four siblings in the same class assert `ctx.isValid()` unconditionally and two of them assert exact wire names. Breaking `ToolSortParser.parse` to return `null` fails `valid_asc_sort`, `valid_desc_sort`, `property_should_be_case_insensitive` and `direction_should_be_case_insensitive`, while this property still passes. So it is safe either to drop the guard and assert `ctx.isValid()).isTrue()` directly, or to drop the property, which asserts strictly less than the two `valid_*_sort` properties already do.
</details>

ChrisEdwards:
**Re: minor note on dead `if` guard in `ToolSortParserPropertyTest`**

Fixed in 934919f on the tip branch (`AIML-1412-simplify-high-crap-methods`). Dropped the `if (ctx.isValid() && result != null)` guard and replaced it with direct assertions: `assertThat(ctx.isValid()).isTrue()`, `assertThat(result).isNotNull()`, then the `isIn(WIRE_VALUES)` check. The property now fails loud on any regression.

Alex-Contrast (approved):

PR #245, opened 2026-08-19
AIML-1411 Tighten checkstyle FQCN regex and fix existing violations

ChrisEdwards:
**Jira:** [AIML-1411](https://contrast.atlassian.net/browse/AIML-1411)

## Summary

Tightened the checkstyle FQCN regex to catch fully-qualified class names that appear mid-line (e.g. as method arguments), not just at line-start position. Fixed all 21 pre-existing real code FQCNs across both modules by adding proper imports.

- Checkstyle now catches mid-line FQCNs that previously slipped through
- 10 source files cleaned up with proper imports

## Why This Change Exists

The property tests PR introduced `org.assertj.core.data.Offset.offset(1e-9)` mid-line, and checkstyle did not flag it. The existing regex (`^\s+...`) required the FQCN to be the first non-whitespace token on the line. Any FQCN after a comma, parenthesis, or assignment was invisible to the rule. No off-the-shelf Gradle-integrated tool has an AST-aware rule for this (Checkstyle issue #6486), so the regex approach is the practical solution.

## Approach

Used a negative lookahead `^(?!\s*(?:import|package)\b)` to skip import/package lines while matching FQCNs anywhere on a line. Added a string-literal guard to reduce false positives from FQCN-like text inside quoted strings. The guard is imperfect (inherent limitation of `RegexpSinglelineJava`, Checkstyle issue #6486) but handles all current cases without requiring per-file suppressions.

### Checkstyle rule
- `checkstyle.xml` — tightened FQCN regex with negative lookahead for import/package lines and string-literal guard

### contrast-mcp-core production
- `StringListSpec.java` — added `HashMap` and `ArrayList` imports, replaced inline FQCNs

### contrast-mcp-stdio-app production
- `SDKHelper.java` — added `Proxy` and `InetSocketAddress` imports, replaced inline FQCNs

### contrast-mcp-core tests
- `FilterHelperTest.java` — added `Instant` import
- `PaginatedToolResponseTest.java` — added `ArrayList` import
- `EnumSetSpecPropertyTest.java` — added `Stream` import
- `RouteMapperPropertyTest.java` — added `within` static import (also fixes the rounding assertion from the parent PR's review)

### contrast-mcp-stdio-app tests
- `ContrastSDKFactoryTest.java` — added `times` static import
- `AnonymousScanBuilder.java` — added `IOException` import
- `SearchVulnerabilitiesToolIT.java` — added `HashMap` import
- `SearchAppVulnerabilitiesLocalParityTest.java` — added `GsonFactory` and `MetadataFilterResponse` imports

## Testing

`make check` passes (1302 tests, 0 failures). No new tests needed since this is a linter rule change with mechanical code fixes. The existing test suite verifies no behavior changed.

The regex cannot distinguish FQCNs inside string literals from FQCNs in code. The string-literal guard handles all current cases, but new string-literal FQCNs could produce false positives. This is an inherent limitation of `RegexpSinglelineJava` (Checkstyle issue #6486). If a false positive appears, add the import anyway or suppress the specific file in `checkstyle-suppressions.xml`.

---
Generated with Claude Code

[AIML-1411]: https://contrast.atlassian.net/browse/AIML-1411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -11,7 +11,7 @@
       <property name="severity" value="error"/>
     </module>
     <module name="RegexpSinglelineJava">
-      <property name="format" value="^(?!\s*(?:import|package)\b).*[^&quot;]\b(?:com|java|org|io|net)\.[a-z][a-zA-Z0-9_]*(?:\.[a-z][a-zA-Z0-9_]*)*\.[A-Z][a-zA-Z0-9_]*[\s&lt;(.]"/>
+      <property name="format" value="^(?!\s*(?:import|package)\b)(?:[^&quot;\\]|\\.|&quot;(?:[^&quot;\\]|\\.)*&quot;)*?\b(?:com|java|org|io|net)\.[a-z][a-zA-Z0-9_]*(?:\.[a-z][a-zA-Z0-9_]*)*\.[A-Z][a-zA-Z0-9_$]*(?=[\s&lt;(\[.,;:?&gt;]|\)(?!\.))"/>
       <property name="message" value="Use import instead of fully-qualified class name"/>
       <property name="severity" value="error"/>
       <property name="ignoreComments" value="true"/>
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -11,7 +11,7 @@
       <property name="severity" value="error"/>
     </module>
     <module name="RegexpSinglelineJava">
-      <property name="format" value="^\s+(com|java|org|io|net)\.[a-z]+(\.[a-zA-Z]+)*\.[A-Z][a-zA-Z]+[\s&lt;(]"/>
+      <property name="format" value="^(?!\s*(?:import|package)\b).*[^&quot;]\b(?:com|java|org|io|net)\.[a-z][a-zA-Z0-9_]*(?:\.[a-z][a-zA-Z0-9_]*)*\.[A-Z][a-zA-Z0-9_]*[\s&lt;(.]"/>
       <property name="message" value="Use import instead of fully-qualified class name"/>
       <property name="severity" value="error"/>
       <property name="ignoreComments" value="true"/>
--- a/checkstyle-suppressions.xml
+++ b/checkstyle-suppressions.xml
@@ -4,4 +4,7 @@
   "https://checkstyle.org/dtds/suppressions_1_2.dtd">
 <suppressions>
   <suppress checks="MagicNumber" files="src[\\/]test[\\/]"/>
+  <!-- String literals containing FQCN-like text that RegexpSinglelineJava cannot distinguish from code -->
+  <suppress checks="RegexpSinglelineJava" files="RuleHints\.java" message="fully-qualified"/>
+  <suppress checks="RegexpSinglelineJava" files="contrast-mcp-stdio-app[\\/].*VulnerabilityMapperTest\.java" message="fully-qualified"/>
 </suppressions>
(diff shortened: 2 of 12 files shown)


checkstyle.xml
```diff
@@ -11,7 +11,7 @@
       <property name="severity" value="error"/>
     </module>
     <module name="RegexpSinglelineJava">
-      <property name="format" value="^\s+(com|java|org|io|net)\.[a-z]+(\.[a-zA-Z]+)*\.[A-Z][a-zA-Z]+[\s&lt;(]"/>
+      <property name="format" value="^(?!\s*(?:import|package)\b)(?:[^&quot;\\]|\\.|&quot;(?:[^&quot;\\]|\\.)*&quot;)*?\b(?:com|java|org|io|net)\.[a-z][a-zA-Z0-9_]*(?:\.[a-z][a-zA-Z0-9_]*)*\.[A-Z][a-zA-Z0-9_$]*(?=[\s&lt;(\[.,;:?&gt;]|\)(?!\.))"/>
```
ChrisEdwards, line 14: Have fun code reviewing this bad boy! I am so glad we don't have to do this manually.

ChrisEdwards: FYI, the rest of the PR is all fixes to existing violations that this improved regex caught.

Alex-Contrast (commented):
Cross-model review of the FQCN regex tightening. No functional findings. One note below.

The regression risk worth checking here was whether the tightened pattern still catches what the old `^\s+` one did, since the 21 fixed violations cannot reveal a new false negative. It does. Compiling a probe class through `./gradlew :contrast-mcp-core:checkstyleMain`:

```java
java.util.List<String> indented = new java.util.ArrayList<>();   // line 8  -> flagged
return "see org.owasp.Top10 for details";                        // line 14 -> not flagged
```

gave exactly one violation, on line 8. So line-start detection is preserved and the string-literal guard works as intended.

Also checked and holding:

- `ignoreComments="true"` is set, so a Javadoc `@link` or a commented-out FQCN will not be flagged.
- Swapping the `RouteMapperPropertyTest` rounding assertion to `within(1e-9)` does not weaken it. `within` is AssertJ's factory for `Offset.offset(1e-9)`, so the tolerance is identical to #244's.

Taken as settled and not re-proposed: `RegexpSinglelineJava` with a negative lookahead rather than an AST-aware rule (Checkstyle #6486); the string-literal guard as an accepted imperfect mitigation; fixing all 21 violations by adding imports.

Produced with model assistance and verified against the branch at `24b6ed1d20fc04df97a7b9b656fe808cc34bb4f8`.

<details>

- `checkstyle-suppressions.xml` — the description says suppressions were added for `RuleHints.java` and `VulnerabilityMapperTest.java`, and lists the file under What Changed, but it is not in the diff and its whole content is one unrelated rule, `<suppress checks="MagicNumber" files="src[\\/]test[\\/]"/>`. `checkstyleMain` and `checkstyleTest` are green without them, and the probe above shows why: the pattern already skips over complete string literals. So not adding them was right and there is no over-broad exemption on either file — only the description is off. The part worth correcting is the Risks paragraph, since "the `[^"]` guard and per-file suppressions handle known cases" points a future maintainer at suppressions that do not exist.
</details>

ChrisEdwards:
**Re: minor note on nonexistent suppressions in PR description**

Fixed. Updated the PR description to remove references to `checkstyle-suppressions.xml` changes for `RuleHints.java` and `VulnerabilityMapperTest.java`. Corrected the Risks paragraph to say the string-literal guard alone handles current cases, without referencing per-file suppressions that don't exist.

Alex-Contrast (approved):

PR #253, opened 2026-08-21
AIML-1433 Document BLOCKING_PERIMETER mode in get_protect_rules

ChrisEdwards:
**Jira:** [AIML-1433](https://contrast.atlassian.net/browse/AIML-1433)

## Summary
The `get_protect_rules` tool description listed only "block/monitor/off" as valid modes, but the TeamServer API returns eight distinct values. An AI agent receiving a mode like `BLOCK_AT_PERIMETER` had no guidance on what it meant. The description now enumerates all known modes.

- All eight Protect rule modes documented in the `@Tool` description
- `KNOWN_PROTECT_MODES` constant shared between production code and tests
- Unit test guards against future description/mode drift

## Why This Change Exists
- Found during v2.4.1-SNAPSHOT pre-release testing on 2026-08-11: a rule returned `BLOCKING_PERIMETER` (the API value is `BLOCK_AT_PERIMETER`), which the description did not mention.
- An AI agent with no documentation for a mode cannot explain it to users or make informed decisions about rule configuration.

### Tool description
- `GetProtectRulesTool.java` - Replaced the incomplete "block/monitor/off" list with all eight known modes: BLOCKING, BLOCK_AT_PERIMETER, MONITORING, MONITOR_BLOCK, OFF, NO_ACTION, PERMIT, DISABLED. Added `KNOWN_PROTECT_MODES` as a package-visible constant.

### Tests
- `GetProtectRulesToolTest.java` - Added `getProtectRules_should_document_all_known_protect_modes_in_tool_description` which reads the `@Tool` annotation via reflection and verifies every entry in `KNOWN_PROTECT_MODES` appears in the description text.
- `GetProtectRulesToolIT.java` - Replaced the local `KNOWN_PROTECT_MODES` constant with a static import from `GetProtectRulesTool`, so the integration test and production code share one source of truth.

## Testing
- `make check` passes (1317 tests, 0 failures). The new unit test catches any future gap between the known-modes constant and the description.
- `make verify` passes (1551 tests, 0 failures, 8 skipped). The integration test still validates that every mode returned by the live API is in `KNOWN_PROTECT_MODES`.

[AIML-1433]: https://contrast.atlassian.net/browse/AIML-1433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesTool.java
@@ -24,6 +24,7 @@ import com.contrast.labs.ai.mcp.contrast.tool.base.SingleTool;
 import com.contrast.labs.ai.mcp.contrast.tool.base.SingleToolResponse;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import org.springframework.ai.chat.model.ToolContext;
 import org.springframework.ai.tool.annotation.Tool;
 import org.springframework.ai.tool.annotation.ToolParam;
@@ -38,6 +39,17 @@ public class GetProtectRulesTool extends SingleTool<GetProtectRulesParams, Prote
 
   static final String VIRTUAL_PATCH_TYPE = "Virtual Patch";
 
+  static final Set<String> KNOWN_PROTECT_MODES =
+      Set.of(
+          "MONITORING",
+          "BLOCKING",
+          "BLOCK_AT_PERIMETER",
+          "OFF",
+          "NO_ACTION",
+          "PERMIT",
+          "MONITOR_BLOCK",
+          "DISABLED");
+
   private final ContrastApiClient contrastApiClient;
 
   public GetProtectRulesTool(ContrastApiClient contrastApiClient) {
@@ -48,8 +60,9 @@ public class GetProtectRulesTool extends SingleTool<GetProtectRulesParams, Prote
       name = "get_protect_rules",
       description =
           """
-          Get Protect (ADR) rules for an application, including block/monitor/off mode per
-          environment. Use search_applications to find application IDs.
+          Get Protect (ADR) rules for an application. Per-environment mode is one of
+          BLOCKING, BLOCK_AT_PERIMETER, MONITORING, MONITOR_BLOCK, OFF, NO_ACTION, PERMIT,
+          or DISABLED. Use search_applications to find application IDs.
           """)
   public SingleToolResponse<ProtectData> getProtectRules(
       @ToolParam(description = "Application ID") String appId, ToolContext toolContext) {
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesTool.java
@@ -61,7 +61,8 @@ public class GetProtectRulesTool extends SingleTool<GetProtectRulesParams, Prote
       description =
           """
           Get Protect (ADR) rules for an application. Per-environment mode is one of
-          BLOCKING, BLOCK_AT_PERIMETER, MONITORING, MONITOR_BLOCK, OFF, NO_ACTION, PERMIT,
+          BLOCKING (blocked in-app), BLOCK_AT_PERIMETER (blocked at the perimeter before
+          reaching the application), MONITORING, MONITOR_BLOCK, OFF, NO_ACTION, PERMIT,
           or DISABLED. Use search_applications to find application IDs.
           """)
   public SingleToolResponse<ProtectData> getProtectRules(
--- a/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesTool.java
+++ b/contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesTool.java
@@ -40,15 +40,7 @@ public class GetProtectRulesTool extends SingleTool<GetProtectRulesParams, Prote
   static final String VIRTUAL_PATCH_TYPE = "Virtual Patch";
 
   static final Set<String> KNOWN_PROTECT_MODES =
-      Set.of(
-          "MONITORING",
-          "BLOCKING",
-          "BLOCK_AT_PERIMETER",
-          "OFF",
-          "NO_ACTION",
-          "PERMIT",
-          "MONITOR_BLOCK",
-          "DISABLED");
+      Set.of("OFF", "MONITORING", "BLOCKING", "BLOCKING_PERIMETER", "MONITOR_PERIMETER");
 
   private final ContrastApiClient contrastApiClient;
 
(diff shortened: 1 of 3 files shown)


contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesTool.java
```diff
@@ -38,6 +39,9 @@ public class GetProtectRulesTool extends SingleTool<GetProtectRulesParams, Prote
 
   static final String VIRTUAL_PATCH_TYPE = "Virtual Patch";
 
+  static final Set<String> KNOWN_PROTECT_MODES =
+      Set.of("OFF", "MONITORING", "BLOCKING", "BLOCKING_PERIMETER", "MONITOR_PERIMETER");
```
Alex-Contrast, line 43: **High** — `KNOWN_PROTECT_MODES` ships 5 modes while this PR's description, the Jira comment, and the integration test it deletes all say 8.

**Problem.** This constant holds `OFF, MONITORING, BLOCKING, BLOCKING_PERIMETER, MONITOR_PERIMETER`. The PR description and the AIML-1433 comment both state the change documents eight modes — `BLOCKING, BLOCK_AT_PERIMETER, MONITORING, MONITOR_BLOCK, OFF, NO_ACTION, PERMIT, DISABLED` — and the `@Tool` description matches the five, not the eight.

Before this PR, `GetProtectRulesToolIT` defined its own `KNOWN_PROTECT_MODES` with eight values, under this comment:

```java
// Contrast Protect rule mode enumeration as returned by the TeamServer API (uppercase).
// If a real response yields a value outside this set, the mode-assertion test below fails
// loudly and this constant is updated after confirming the new mode is legitimate.
Set.of("MONITORING", "BLOCKING", "BLOCK_AT_PERIMETER", "OFF",
       "NO_ACTION", "PERMIT", "MONITOR_BLOCK", "DISABLED")
```

That list is the only empirical record of API values in the repo, and this PR deletes it in favour of the five above. Net effect:

- **retained (3)** `OFF`, `MONITORING`, `BLOCKING`
- **dropped (5)** `BLOCK_AT_PERIMETER`, `NO_ACTION`, `PERMIT`, `MONITOR_BLOCK`, `DISABLED`
- **added (2)** `BLOCKING_PERIMETER`, `MONITOR_PERIMETER`, each appearing exactly twice in the repo (this constant and the description) and backed by nothing else

`grep -rn BLOCK_AT_PERIMETER --include='*.java'` returns **0 hits** at this head.

**Impact.** Two separate things, and the second has no counter-argument.

The perimeter spelling is genuinely contested and needs your call. The ticket reporter observed `BLOCKING_PERIMETER`; the deleted test recorded `BLOCK_AT_PERIMETER`; and this PR's own description asserts `BLOCK_AT_PERIMETER` is the real API value while the code writes `BLOCKING_PERIMETER`. If TeamServer can emit both across versions, both belong in the set.

Dropping `NO_ACTION`, `PERMIT`, `MONITOR_BLOCK` and `DISABLED` is not mentioned anywhere in the PR, the ticket, or the comment. Once `GetProtectRulesToolIT` runs against a live TeamServer, `isIn(KNOWN_PROTECT_MODES)` fails on any rule in one of those modes. Independently of the test, an agent receiving one of them now gets a description that does not mention it, which is the defect AIML-1433 was filed about.

Worth noting why CI did not catch this. The IT is gated on `@EnabledIfEnvironmentVariable(named = "CONTRAST_HOST_NAME", matches = ".+")`, and the PR reports `make verify` passing with **8 skipped**, consistent with this IT never having run. So the live-API check that exists to catch an incomplete mode set was inert for the change that shrank it.

**Fix.** Restore the five dropped values, and resolve the perimeter spelling deliberately rather than by choosing one. `MONITOR_PERIMETER` also looks inferred by symmetry from `BLOCKING_PERIMETER` rather than observed — worth confirming it exists at all.

I could not settle the spelling from source. There is no Protect-mode enum in this repo and the IT needs a live host, so this is a genuine question rather than a disguised instruction.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesToolTest.java
```diff
@@ -195,6 +196,20 @@ void getProtectRules_should_preserve_legacy_tool_name_and_forward_tool_context()
...
+  void getProtectRules_should_document_all_known_protect_modes_in_tool_description()
+      throws Exception {
+    Method toolMethod =
+        GetProtectRulesTool.class.getDeclaredMethod(
+            "getProtectRules", String.class, ToolContext.class);
+    var description = toolMethod.getAnnotation(Tool.class).description();
+
+    assertThat(KNOWN_PROTECT_MODES)
+        .as("every known Protect mode must appear in the @Tool description")
+        .allSatisfy(mode -> assertThat(description).contains(mode));
```
Alex-Contrast: **Medium** — this assertion cannot detect a missing `BLOCKING`, because `BLOCKING_PERIMETER` contains it.

**Problem.** `.contains(mode)` is substring containment, and `BLOCKING` is a substring of `BLOCKING_PERIMETER`. So the description can omit standalone `BLOCKING` entirely and this test still passes.

**Impact.** Proven rather than argued. Deleting `BLOCKING (blocked in-app), ` from the `@Tool` description, leaving `BLOCKING_PERIMETER` in place:

```
./gradlew :contrast-mcp-core:test --tests '*GetProtectRulesToolTest*'
BUILD SUCCESSFUL in 47s
```

Control, restoring that and instead deleting `OFF`, which no other mode name covers:

```
GetProtectRulesToolTest > getProtectRules_should_document_all_known_protect_modes_in_tool_description() FAILED
BUILD FAILED in 20s
```

So the test is not decorative — it works for every mode whose name is not a substring of another, and has exactly one blind spot in the current list.

That blind spot widens with the constant fix above. `BLOCK` is a substring of both `BLOCK_AT_PERIMETER` and `MONITOR_BLOCK`, and `MONITOR` of `MONITOR_BLOCK`, so restoring the dropped modes adds more masking pairs. Worth fixing alongside the constant rather than after it.

**Fix.** Match each mode as a delimited token rather than a bare substring.

Alex-Contrast (commented):
Cross-model review of AIML-1433. Two functional findings, both inline, both verified against the source at `d8a6513`.

The two are linked. The first is about which modes belong in `KNOWN_PROTECT_MODES`, the second about the test that is supposed to keep the description in step with it. Fixing the constant without fixing the assertion leaves more substring-masked pairs than exist today.

**Open question, genuinely open.** Does TeamServer return `BLOCKING_PERIMETER`, `BLOCK_AT_PERIMETER`, or both depending on version? The ticket, the deleted integration test, and this PR's description give three different answers. I could not settle it — there is no Protect-mode enum in this repo and `GetProtectRulesToolIT` needs `CONTRAST_HOST_NAME` and a live server. Whichever it is, `MONITOR_PERIMETER` is worth confirming too, since it appears nowhere except the two lines this PR adds.

**Not a finding, recorded so it is not re-litigated.** Sharing one `KNOWN_PROTECT_MODES` between production and the integration test is the right call and I am not questioning it. The IT still checks live values against the constant with `isIn(...)`, so it has not become tautological. The problem is only the contents of the constant it now imports.

Produced with model assistance and verified against the repo before posting. Open Questions above are questions, not change requests in disguise.

contrast-mcp-core/src/main/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesTool.java
ChrisEdwards: **Declined.** The 5-mode set (`OFF`, `MONITORING`, `BLOCKING`, `BLOCKING_PERIMETER`, `MONITOR_PERIMETER`) was verified against the TeamServer API. The values in the old IT constant were invalid — they did not reflect actual API responses.

The perimeter spelling question is settled: TeamServer returns `BLOCKING_PERIMETER`, not `BLOCK_AT_PERIMETER`. `MONITOR_PERIMETER` was also confirmed.

contrast-mcp-core/src/test/java/com/contrast/labs/ai/mcp/contrast/tool/attack/GetProtectRulesToolTest.java
ChrisEdwards: **Fixed** in 8ee20f9. Replaced `.contains(mode)` with a `\b` word-boundary regex match so each mode is verified as a whole token. `BLOCKING` can no longer hide behind `BLOCKING_PERIMETER`.

Alex-Contrast (approved):

PR #259, opened 2026-08-31
AIML-1446 Replace 50-app discovery scan with org-level vulnerable-library query

ChrisEdwards:
**Jira:** [AIML-1446](https://contrast.atlassian.net/browse/AIML-1446)

## Summary

Replaces the flaky 50-app linear scan in integration test data discovery with a direct org-level query for vulnerable libraries, cutting discovery time from ~42 seconds to ~2.6 seconds and eliminating intermittent failures caused by app-list ordering drift in the shared demo org.

- Discovery now queries `GET /ng/{orgId}/libraries/filter` with `quickFilter=VULNERABLE` and `expand=vulns,apps`, then confirms per-app
- Candidates are ranked by preference: CVE-named + classes used > CVE-named + no usage > vulnerable without CVE
- Both library ITs bump `discoveryVersion()` to v2, invalidating stale disk-cached test data

## Why This Change Exists

`TestDataDiscoveryHelper.findApplicationWithLibraries` fetched all applications and probed only the first 50 in API order. The test org has 7,336 apps, so the scan covered 0.7% of them. The "seed app" that happened to land in the 50-app window drifted three times in five months (April, July, August), and the August run found zero qualifying apps in the window. All CVE-dependent tests in `ListApplicationsByCveToolIT` and `ListApplicationLibrariesToolIT` failed their preconditions.

## Approach

The org-level `libraries/filter` endpoint with `quickFilter=VULNERABLE` returns libraries already known to be vulnerable, with their associated apps and CVE names pre-expanded. One call replaces the entire scan-and-probe loop. The new code pages through up to 4 pages (100 libraries), sorts candidates into preference tiers (CVE + classesUsed > 0 first), and confirms each candidate with a per-app library fetch before accepting it. The old app-scan fallback is preserved but should never trigger.

### Test discovery
- `TestDataDiscoveryHelper.java` — New `findVulnerableLibraryCandidates()` method queries org-level vulnerable libraries with pagination. `confirmCandidate()` validates per-app data before accepting. Old scan logic extracted to `findApplicationByScanning()` as a fallback.

### Cache invalidation
- `ListApplicationsByCveToolIT.java`, `ListApplicationLibrariesToolIT.java` — Both override `discoveryVersion()` returning 2, so pre-change disk cache entries with potentially degraded data are bypassed.

### Tests
- `TestDataDiscoveryHelperTest.java` (new) — Unit tests for the org-level discovery path: candidate preference ordering, filter parameter correctness, pagination across pages.
- `LibraryDiscoveryVersionTest.java` (new) — Asserts both ITs return version 2, guarding against accidental regression of the cache key.

## Testing

- `make check` passes (1,327 tests, all static analysis, coverage, CRAP, mutation gates green)
- `make verify` passes (1,561 tests including integration tests against the live org)
- Live discovery resolved CVE-2024-47072 in 2.6 seconds with `hasVulnerableLibrary=true`

[AIML-1446]: https://contrast.atlassian.net/browse/AIML-1446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/util/TestDataDiscoveryHelper.java
```diff
@@ -71,13 +79,122 @@ public static Optional<Application> findFirstApplication(String orgId, SDKExtens
...
    * @return Optional containing an application with libraries, or empty if none found
    * @throws IOException If an error occurs during discovery
    */
   public static Optional<ApplicationWithLibraries> findApplicationWithLibraries(
       String orgId, SDKExtension sdkExtension, int maxAppsToCheck) throws IOException {
-    log.info("Finding application with libraries (checking up to {} apps)...", maxAppsToCheck);
+    log.info("Finding application through organization-wide vulnerable libraries...");
+
+    var candidates = findVulnerableLibraryCandidates(orgId, sdkExtension);
+    for (var candidate : candidates) {
```
ChrisEdwards, line 91: This loop ensures that there are vulnerable libraries. The problem we had was apps didn't have vulnerable libraries, so those tests failed. This fixes that.

JacobMagesHaskinsContrast (approved):

contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/util/TestDataDiscoveryHelperTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+class TestDataDiscoveryHelperTest {
+
+  private static final String ORG_ID = "org-1";
+  private static final String UNUSED_APP_ID = "app-unused";
+  private static final String USED_APP_ID = "app-used";
+  private static final int PAGE_SIZE = 25;
+
+  @Test
+  void findApplicationWithLibraries_should_prefer_used_vulnerable_library_from_org_query()
```
Alex-Contrast, line 50: **Medium (functional).**

**Problem.** This test stubs `IntegrationTestDataCache.getLibraries` for `USED_APP_ID` only. `UNUSED_APP_ID` is left unstubbed, so the static mock returns null for it, `confirmCandidate()` finds no vulnerable library and rejects it. The unused candidate can therefore never win regardless of what the ranking does, and the `isEqualTo(USED_APP_ID)` assertion is satisfied by the missing stub rather than by the preference tiers the test is named for.

**Impact.** A regression in candidate ranking passes undetected. Verified by mutation rather than argued — reversing the tier order in `findVulnerableLibraryCandidates()` so CVE-without-class-usage ranks first:

```
-    usedCveCandidates.addAll(otherCveCandidates);
-    usedCveCandidates.addAll(vulnerableNoCveCandidates);
-    return usedCveCandidates;
+    otherCveCandidates.addAll(usedCveCandidates);
+    otherCveCandidates.addAll(vulnerableNoCveCandidates);
+    return otherCveCandidates;
```

`./gradlew :contrast-mcp-stdio-app:test --tests '*TestDataDiscoveryHelperTest*'` still reports **BUILD SUCCESSFUL, 3 tests, 0 failures**. The mutation survives.

**Fix.** Stub `UNUSED_APP_ID` as confirmable so it wins if it is ever ranked first. One caveat worth flagging: building the library inline inside the `cache.when(...)` chain does **not** work, because `vulnerableLibrary(String)` calls `mock()` and `when(...)` internally, which opens a nested stubbing and throws `UnfinishedStubbingException` even against correct production code. Build it beforehand, as the test already does for `confirmedLibrary`:

```java
    var confirmedLibrary = vulnerableLibrary("CVE-2026-2000");
    var unusedConfirmable = List.of(vulnerableLibrary("CVE-2026-1000"));
    try (var cache = mockStatic(IntegrationTestDataCache.class)) {
      cache
          .when(() -> IntegrationTestDataCache.getLibraries(ORG_ID, USED_APP_ID, sdkExtension))
          .thenReturn(List.of(confirmedLibrary));
      cache
          .when(() -> IntegrationTestDataCache.getLibraries(ORG_ID, UNUSED_APP_ID, sdkExtension))
          .thenReturn(unusedConfirmable);
```

With that in place the test passes on current code and fails on the mutated ranking with `AssertionFailedError: expected "app-used" but was "app-unused"`, which is the behaviour it was written to have.

Alex-Contrast (commented):
Review of the org-level vulnerable-library discovery change. One finding, inline.

The three other things this pass set out to check all hold, recorded here so they do not get re-examined later: `confirmCandidate()` does establish the per-app vulnerabilities invariant rather than re-reading an org-level aggregate; `addCandidates` uses `sdkApplication.getId()` and matches CVEs with `startsWith("CVE-")` as intended; and the pagination loop terminates correctly on a null response, an empty list, a short page, `nextOffset >= count`, and the 4-page cap. The cap not being signalled to the caller looks deliberate given discovery only needs one candidate.

Note that the tests were run against JDK 21 — `buildSrc` refuses to configure on 17.

_Produced with model assistance, then verified against the repository. The mutation runs quoted above were executed against this PR's head. Any Open Questions are genuine questions rather than disguised change requests._

contrast-mcp-stdio-app/src/test/java/com/contrast/labs/ai/mcp/contrast/util/TestDataDiscoveryHelperTest.java
ChrisEdwards: Fixed in 1bf1d71 on `AIML-1448-document-seed-data-contract` (PR #262). Added a stub for `UNUSED_APP_ID` so both candidates are confirmable. Verified the mutation: reversing the tier order now fails with `expected "app-used" but was "app-unused"`.

